PR #5, opened 2015-02-24
comma missing from analog json output

patrickfish:
Add comma to json output.

Before:
```json
{
"freq": 8.53225e+08,
"num": 6,
"talkgroup": 12944
"mode": "analog"
}
```

After:
```json
{
"freq": 8.53225e+08,
"num": 6,
"talkgroup": 12944,
"mode": "analog"
}
```


robotastic:
Thanks for catching this! I had the same problem with the Digital one and forgot to change it here.

PR #15, opened 2015-08-18
Add some logging to talkgroups.cc

lachesis:
The ChanList.csv parser is pretty brittle. This at least adds some logging to help troubleshoot why talkgroups are not found, and also makes it work correctly with all the standard line ending formats.


robotastic:
Awesome! Thanks - Do you have trunk-recorder up and running?

lachesis:
Yep! It was actually pretty painless on Arch. I was thinking about trying
to clone OpenMHZ for Oakland next.

-Eric (mobile)

PR #16, opened 2015-08-18
Strip trailing whitespace

lachesis:
Just a small style change. My editor gets mad about the trailing whitespace, and I agree that it's a bit wacky. If you don't like this, feel free to reject, but I'm going to start basing my changes on it.


robotastic:
Cleaner is better!

PR #17, opened 2015-08-21
Update README.md

brendanvg:
Under GR-DSD installation, changed lbsndfile to lbsndfile1  
check it -> https://launchpad.net/ubuntu/+source/libsndfile

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ If you are on OSX, the [MacPorts](https://gnuradio.org/redmine/projects/gnuradio
 **GR-DSD**
 
 I made a fork of this code to allow for more statistics to be collected and the also make usre multiple copies can run at once. It has a few dependencies:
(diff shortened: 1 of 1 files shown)


robotastic:
Awesome! Good catch, thanks!

brendanvg:
hey Luke!
First off....huge respect for writing the trunk recorder and
smartnet-player...they're exactly the tools I need and it seems like they
are pretty well-equipped....so, I've got everything configured, but how do
I click "record" ???  and how do I listen to make sure I'm recording
everything I want to? Do I have to go through smartnet-player???

brendanvg:
ok...so I see the "recorder" file...that seemed to work...what about
listening to what I'm getting?  any recommended methods to testing how much
data on the trunked system I've actually successfully received?

PR #20, opened 2015-08-23
Clean up logging with BOOST_LOG_TRIVIAL

lachesis:
I replaced (most) instances of std::cout with BOOST_LOG_TRIVIAL.

I also removed the unneeded std::endl from these calls. There is probably
still some work to be done to clean up logging, but this is a good
start.


robotastic:
Awesome! Thanks

PR #37, opened 2016-03-19
Fix PPM in init logging

hatchetation:
Was trying to figure out why my `ppm` directive wasn't taking effect... Turns out it's just a logging bug.


robotastic:
doh! Good catch, thanks for fixing it! Let me know how accurate the PPM approach is. I am using the underlying code in the OsmoComm drivers.

PR #42, opened 2016-05-01
Stop a couple of warn_unused_result warnings

dreinhold:
2 system calls did not have their return code assigned to anything. 

/home/dylan/radio/dylan-trunk-recorder/main.cc: In function ‘void unit_check()’:
/home/dylan/radio/dylan-trunk-recorder/main.cc:507:30: warning: ignoring return value of ‘int system(const char*)’, declared with attribute warn_unused_result [-Wunused-result]
         system(shell_command);

/home/dylan/radio/dylan-trunk-recorder/call.cc: In member function ‘void Call::end_call()’:
/home/dylan/radio/dylan-trunk-recorder/call.cc:75:38: warning: ignoring return value of ‘int system(const char*)’, declared with attribute warn_unused_result [-Wunused-result]
                 system(shell_command);

### DIFF
--- a/call.cc
+++ b/call.cc
@@ -72,7 +72,7 @@ void Call::end_call() {
                 }
                 sprintf(shell_command,"./encode-upload.sh %s > /dev/null 2>&1 &", this->get_filename());
                 this->get_recorder()->deactivate();
-                system(shell_command);
+                int rc = system(shell_command);
             }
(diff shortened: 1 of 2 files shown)


robotastic:
Good catch, I have been too lazy to fix this ...

PR #47, opened 2016-05-13
Analog calls were being flaged as TDMA

dreinhold:
message.tdma was never definded for an analog call, now defaulted to false

  Added .so files to gitignore


robotastic:
Good Catch!

PR #52, opened 2016-06-07
Add System ID as info output when it is decoded.

dreinhold:
Example output:
  [2016-06-06 21:50:02.360250] [0x00007f92f1b0f780] [info]    Decoding System ID 4A6

### DIFF
--- a/main.cc
+++ b/main.cc
@@ -447,7 +447,13 @@ void add_control_channel(double control_channel) {
     }
 }
 
-
+void current_system_id(int sysid) {
+    static int active_sysid = 0;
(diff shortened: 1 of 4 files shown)


robotastic:
Looks good - I am going to try and add in some more P25 SysIds from messages.

What are you using SystemID for?

dreinhold:
> What are you using SystemID for?

I use it just as a sanity check that I started recorder on the right system. Mainly after testing a change I switch over to a P25 system from the smartnet system I normally monitor. 
Also if the system is slow and there is no voice traffic you can still see that you are decoding the control chanel properly.

Issue #53, opened 2016-06-08
Smartnet Radio ID Issue


dreinhold:
I just noticed my last commit de5942c9d was grabbing the radio id from the last message.source.
The issue is message.source has its least significant byte removed (well set to 0) as that is the status byte when it's the channel id.
I have a patch I'll submit tomorrow, just wanted to open this if anyone one else is using the latest version and notices some duplicate radio ids on a smartnet system.

robotastic:
I just merged in the change. Thanks for submitting, and catching this!

PR #54, opened 2016-06-10
Save all 4 bytes from the address into last address

dreinhold:
We were setting the last byte to a 0, the last byte when we get a talkgroup
is the type of call. But when its a radio ID all 4 bytes are used.

This will correct issue #53


robotastic:
Good catch!

PR #57, opened 2016-06-16
Capture Astro Digital talkgroups in smartnet systems

dreinhold:
Include the Astro Grant op code x321
    Exclude those with a status > 8 which are encrypted
    Also flag emergency calls


robotastic:
Can you do a Pull of the latest version and resubmit? It got rid of this line/test:
if (  (address != 56016) && (address != 8176)) {

dreinhold:
Opps. Yeah.

dreinhold:
I pushed an amended commit. Should not back-level anything now..

robotastic:
Awesome! Looks good, thanks for the push!

Issue #55, opened 2016-06-10
rtl device is hardcoded in osmosdr connection


dreinhold:
If you pass a device ID in config.json it has the rtl driver hard coded.
Using airspy works fine as the default, but you can't have another rtlsdr connected or else you cannot pick which one you want to use.

Looking for feedback on a couple options : 

1. Just removed the hard-coded rtl= and then require the user pass rtl=0001 as the device in config.json
2. Add a new hybrid driver named osmosdr-airspy. and change the relevant checks

I am leaning towards 2 (this is how I have been doing it). The reason I'm looking at 2 is gain options are also different for airspy. It has IF, MAX, and LNA gain's available to set.
I have some code that also allows those to be set via config.json for another day..

robotastic:
Hmm... it might actually make more sense to go with option 1 and just have IF, Max and LNA be config setting that can be set. I did something similar for HackRF support. If you have code for option 2 already, lets just go with that. I could do something similar in the future for HackRF.

PR #59, opened 2016-07-01
Allow selecting the SDR device type for the osmosdr driver

dreinhold:
When using mutiple SDRs with the osmosdr driver allow user to explicitly select which model of SDR and serial number.
  For backward compatibility if the first part of device is numeric assume this is a serial number and fail back to using the rtl model.
  This corrects issue #55


robotastic:
Awesome! I will go through it this weekend and get it merged in.

robotastic:
This looks great! Thanks for adding it.

PR #61, opened 2016-07-13
Link WIKI from README

dreinhold:


robotastic:
Good call - I will follow suit with the ARM & docker  install

dreinhold:
Are you running trunk-recordor on a PI?

robotastic:
Not for the DC system, I am using a beefy I7. I played around with a RasPi2
and it could handle a control channel and one recording channel. I think
the Pi3 could be mildly useful for a very small system.

dreinhold:
Ha. Okay, I was trying to figure out how you could manage that.
Yeah I have an i5 chugging away for mine.

PR #93, opened 2017-01-26
Filter out sub-audible data from analog transmissions

dreinhold:


robotastic:
Cool! I am getting some bad analog recordings one of my Smartnet systems. I wonder if it is from this.

PR #94, opened 2017-01-27
Record source radio ids from analog calls

dreinhold:


patrickfish:
I tested this locally with no issues. Sources for analog calls are now being recorded in the json file.

dreinhold:
Luke added it back in already.

PR #96, opened 2017-02-04
Update smartnet bandplans

dreinhold:
Add lookups for 800MHz Standard, Splinter and Rebanded sites

Add new config.json option under system named "bandplan":
800_Standard * This is the base plan if an invalid
             * plan is entered it will fall back to this one.
800_Splinter
800_Reband * This is the default if nothing is set in the config

Used https://github.com/ScanOC/Trunk-Channel-Check/raw/master/MO%20TRS%20Channels.pdf as my source of truth


robotastic:
This is awesome - I integrated it into the Develop branch, and it should go live soon. Thanks for submiting!

PR #111, opened 2017-03-16
Support tab in talkgroup csv parsing and default priority

grutz:
When copying talkgroups from RadioReference usually tabs separate the columns, not commas. This change adds tab support.

Priority is trunk-recorder specific and must be added for granularity of recording. This change provides a default priority of 1 if not included.


robotastic:
Thanks! I have been doing the conversion in Excel.
I will get this merged in this weekend.

robotastic:
Awesome! This looks good. Just merged into the master.

PR #112, opened 2017-04-04
fix markdown formatting

nkwood:
- github had issues displaying some of the markdown due to lack of spaces
 - add source highlighting for some blocks


robotastic:
Awesome! Thanks! I wrote most of these using the Markdown preview on Atom. It may have messed thigns up. That formatting trick it pretty cool, good to know.

PR #116, opened 2017-04-10
Set to -std=gnu++98 to fix compiling with GCC 6.

nkwood:
This should fix https://github.com/robotastic/trunk-recorder/issues/115 and https://github.com/robotastic/trunk-recorder/issues/81, but I have only tested on Ubuntu 16.10 and 17.04.

PR #117, opened 2017-04-11
cmake: add -pthread for arch linux and move -g to debug flags.

nkwood:
Fix for https://github.com/robotastic/trunk-recorder/pull/116#issuecomment-293264737.


robotastic:
Awesome!

PR #118, opened 2017-04-12
docs: update readme to point to new wiki pages

nkwood:
* add additional platforms and compatible SDRs
* add links to install for generic Ubuntu, Arch Linux, and macOS
* add link to building wiki page
* add link to running wiki page
* delete OS X and Ubuntu 16.04 install instructions from README in
favor of links to wiki


robotastic:
I like it! I was meaning to eventually move everything over to the Wiki instead of the readme. Thanks!

PR #123, opened 2017-04-21
update talkgroups file loading and README

nkwood:
update talkgroups file loading
     - fix default priority handling
     - move includes into implementation file
     - adjust formatting and indentation
     - document talkgroup file columns
     - log error on malformed lines in talkgroup file
update readme with RR info and priority default


robotastic:
Looks good - thanks! Out of curiosity, why move the includes from the .h file to the .cc?

PR #124, opened 2017-05-03
Merge assign_recorder and update_recorder

mrtrev:


robotastic:
This is a good idea. I am working off of the status-server branch. I will see if I can make a similar change there.

PR #127, opened 2017-05-15
clean up CMakeLists.txt and op25_repeat includes

nkwood:
- previous fix for Arch linux forced pthreads flag, this lets CMake handle it as needed
- if op25 was installed on system, trunk-recorder headers would pickup system version rather than local and compile would fail if versions were different
- also adds .vscode to .gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,4 @@ ChanList.csv
 .DS_Store
 .idea/
 cmake-build-debug/
+.vscode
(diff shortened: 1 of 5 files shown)


nkwood:
Actually this appears to break some stuff, I'm going to close this until I investigate further.

robotastic:
Actaully - if you are able to rebaseline off of the status-socket branch, that would be aswesome. I tried to consolidate the libraries to make things a little easier to manage.

PR #147, opened 2017-11-29
Re-creating status-socket

rryder:


robotastic:
Can you provide a better description around this change? Are you noticing better audio performance from it?

PR #148, opened 2017-12-09
New config for control channel rate warning

dreinhold:
I always seem to to change this and recompile trunk-recorder each time I pull it down..

New config option controlWarnRate to set the warning threshhold to log the control channel decode message.
Default is 10.
-1 will always log


rosecitytransit:
I've just been commenting out the if statement to get it to always show, but having an official configuration option is even better.

On a related note, it would be ideal to change the `timeDiff` for when things go awry and I need to slightly adjust the antenna.  I've been manually changing it and recompiling.  For then I want the rate updated constantly but otherwise only need it every minute for confirmation that it's working good.  But setting the warning threshold high and presuming that everything is going great in the absence of rate messages can also work.

PR #150, opened 2017-12-11
Have System ID, WACN, NAC output in hex

rosecitytransit:
Hex is what is commonly used for these.  I hope I did this right, haven't tested it yet.  See also #52


robotastic:
This is great!! Thanks for contributing!

Issue #79, opened 2016-12-27
Include a sample output Web page


rosecitytransit:
I should have done this before, but I am attaching a stripped down version of the Web page that I created that acts as an interface to the recorded files.  I know there's https://github.com/robotastic/smartnet-player but I didn't want to deal with a database so instead created a PHP script and HTML that can read through all the audio files in a directory and then play them sequentially using the AUDIO tag.

Note that it will need further customizing to handle details from the JSON files that trunk-recorder is designed to create (I modified it so everything is put in the filenames, using HHMMSS-talkgroup-radio-radio.mp3, a better option may be to have it create per-day delimited log files)

[audioplayer.php](https://github.com/robotastic/trunk-recorder/files/674706/basic.php.txt)
[The live version that I am using is here](http://www.rosecitytransit.org:90/radio/test2.php)

robotastic:
This is great! Thanks for sharing back.

dpknoll:
This webpage is outstanding. Looking forward to using this. I'm trying to get this to work but am receiving a 500 error, more specifically apache tail shows - PHP Parse error:  syntax error, unexpected '<' in /var/www/html/basic.php on line 103. 

I'm new to .php so I'm likely missing something simple here. Any ideas on what I can do to resolve?

robotastic:
I setup a portal that should make it easier to view recordings. You need to be running the latest version of trunk-recorder, because it automatically uploads files.

Goto: https://admin.openmhz.com/ and sign up for an account. After you do that, create a new system and it will give you the code you need to add to a config file.

porcej:
Submitting the signup form (https://admin.openmhz.com/signup) gives a 502 bad gateway error.

robotastic:
Doh! This is still a work in progress. I will figure it out and post back when it is working.

rosecitytransit:
OK, I have gone back and finished the page.  It appears that line 103 was incomplete.

[audioplayer.php.txt](https://github.com/robotastic/trunk-recorder/files/706321/audioplayer.php.txt)

You'll want to put it in the captureDir, where the year folders are or customize the $dir parameter.  You'll also want to fill up the talkgroups array on line 12 replacing the Example1 and Example2 ones with the ones you're recording.

It still does not handle the details in the JSON files (e.g. radio numbers) since I got rid of them and it takes resources to read 100s of little files (as I said, a solution would be a have a per-day log file instead/also).

This was based off of
[...]

PR #151, opened 2017-12-12
Add basic audio player v2 (see #79)

rosecitytransit:
Simple PHP Web page that can list audio files in trunk-recorder's output directories and, using HTML5, sequentially play all ones on a certain day,  Supports filtering by channel.  Note that this does not parse the data in the JSON files.

Version 2 (audioplayer-v2.php) supports getting channel names from ChanList,csv, uses a hopefully faster JavaScript way to select the next call, and other improvements.

Note that trunk-recorder's captureDir should be set to something in your Web site directory (usually /var/www/html) and existing calls moved there, the Web page should be in the same directory as trunk-recorder's year folders, and that to use ChanList,csv, it will need to be readable by your Web server program,

### DIFF
--- a/audioplayer-v2.php
+++ b/audioplayer-v2.php
@@ -50,37 +50,39 @@ else echo "radio calls" ?> on <?php echo substr($m,5)."/".$d."/".substr($m,0,4);
 	var playlist;
 	var currentcall;
 	function init() {
+	playlist = document.getElementById('theplaylist');
+	if (playlist.getElementsByTagName('div')[0]) {
 		player = document.getElementById('audioplayer');
-		playlist = document.getElementById('theplaylist');
 		currentcall = playlist.getElementsByTagName('div')[0];
 		currentcall.style.fontWeight = "bold";
-		if (currentcall.getElementsByTagName('a')[0])
-			player.setAttribute('src',currentcall.getElementsByTagName('a')[0]);
-                //document.getElementById("myBase").href = "<?php echo $dir; ?>";
+		player.setAttribute('src',currentcall.getElementsByTagName('a')[0]);
 		player.volume = 0.2;
 
 		playlist.addEventListener('click',function (e) {
 			//e.preventDefault();
-		if (e.target.parentElement.getElementsByTagName('a')[0]) {
-			currentcall.style.fontWeight = "normal";
-			currentcall = e.target.parentElement;
-			currentcall.style.fontWeight = "bold";
-			player.setAttribute('src',currentcall.getElementsByTagName('a')[0]);
-			player.load();
-			player.play();
-		} }, false);
-
-		player.addEventListener('ended',function () { 
-		if (currentcall.nextSibling && (document.getElementById('continuous').checked == true)) {
-			currentcall.style.fontWeight = "normal";
-			currentcall = currentcall.nextSibling;
-			currentcall.style.fontWeight = "bold";
-			player.setAttribute('src',currentcall.getElementsByTagName('a')[0]);
-			player.load();
-			player.play();
-		}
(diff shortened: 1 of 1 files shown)


robotastic:
Awesome - Thanks! I will create a page in the Wiki

bctrainers:
The myBase base HTML tag seems to be throwing chrome for a good loop.

On first glance, It would appear to not be including the system-name or ID tagged to the system configured. For me, the system name/ID being out putted it 688, the directory structure ends up like this: http://192.168.20.3:9099/2017/12/24/688/ - but the /688/ is not being attached to the URL given of the recorded system.


On further delving into the myBase href, chrome duplicates the URL...

`<base id="myBase" href="2017/12/24/">` is what the browser source shows.

When clicking on the href, I end up at http://192.168.20.3:9099/2017/12/24/2017/12/24/

Which seems quite incorrect.


Attached below is a tiny-directory cut out of todays recordings (more than 10k today already, average is 35k files generated daily per date)...
chrome_2017-12-24_10-27-16


With the audio player being in the root directory of this setup, the fix appears to be this...

`$dir = "/".str_replace("-","/",$m)."/".$d."/";`

And for ones with a naming convention per system, the resolution is this...

Adding `$sysid=688;` below `$desc` variable.
Changing `$dir` to `$dir = "/".str_replace("-","/",$m)."/".$d."/".$sysid."/";`


----


Firefox reports a general failure to play files regardless...

`Cannot play media. No decoders for requested formats: text/html`

bctrainers:
Do note, the above changes got the directory structure to load properly, but it will not parse the directory in full for recording playback or any sort of audio.

rosecitytransit:
I'm not sure where the 688 is coming from.  I'm looking at [trunk-recorder/call.cc](https://github.com/robotastic/trunk-recorder/blob/master/trunk-recorder/call.cc) and the directories should be `"/" << sys->get_short_name() << "/" << 1900 + ltm->tm_year << "/" << 1 + ltm->tm_mon << "/" << ltm->tm_mday;`

Also, the Web page should be put in the directory where the 2017 directory is, not with actual recordings.  What's supposed to happen is PHP takes the date and changes to the directory and then outputs a list of files with no directory information, with that coming from the base href (since it's going to be the same for every single one).

If you try to click on the file links at the end of the rows, do they work?  Those are what the audio player tries to load.

bctrainers:
Apologies, the 688 is the short-name/systemID-name i've set. The directory structure ends up as: SERVER_ROOT/2017/12/25/688/. I'm recording two p25 systems, 688 being the busy one.

For some reason this morning, the files are loading in and all is well except for a slight edit required for the file playback being in a custom directory and i'm smooth sailing. Do note, I am using the @treehouseman branch here on github, hence the odd setup for file structure. I love the ncurses UI on the terminal - can't go wrong with it via a screen session, hopefully will be added to the main branch soon. :)

I have attached my copy, which probably requires some code cleanup. :-)

[audioplayer2a.txt](https://github.com/robotastic/trunk-recorder/files/1585790/audioplayer2a.txt)


I do have one issue with the ALL setting in the drop down, it's setting playback of files in A-Z then date of talk groups rather a date layout of oldest to newest (or even newest to oldest) transmission. Currently poking about with it on file sorting.

chrome_2017-12-25_05-26-23

rosecitytransit:
I haven't looked at that code so am not familiar with how it outputs stuff.  As for the "ALL" setting, the problem is that the talkgroup is the first thing in the file name and that's how the directories get sorted.  I change the code so the time is first, and being able to use a log file would eliminate the issue.

For now, I a solution is to load the file list into an array that can be sorted before outputting it.

PR #158, opened 2017-12-26
Fixed a bug that caused systems to be duplicated.

amlethojalen:
Fixed a bug that causes the systems to be duplicated.


robotastic:
Thanks! I probably botched this during merges

PR #159, opened 2017-12-26
Frequency

amlethojalen:
Removed build warnings from dependent lib builds
Addressed Debug Build warnings (commented out the unused variables, and fixed a few signed/unsigned conversions)


robotastic:
Good call - There is a lot of failed experiments left around that I have to clean up. Thanks!

PR #160, opened 2017-12-27
Basic SmartNet unit ID fix

jdeglopper:
Initial fix for the most common cause of bogus Unit IDs in SmartNet systems.  Accepts 308 and 321 OSW only in position 1, does not process 320 as a voice grant.


robotastic:
Awesome! I am going to go give this a try on my local Smartnet system.

robotastic:
If anyone has wierdness from this, let me know

PR #164, opened 2018-01-10
Several changes

amlethojalen:
You can see all the changes in the individual commits, but the highlevel changes are

- Separated P25 and P25 Conventional call and recorders and 
- optionally hide unknown talkgroups from logs
- optionally hide encrypted talkgroups from logs
- optionally output status as text
- fixed issue with conventional recordings < 1 sec in length form completing
- added talkgroup tags to conventional channels
- udpated the status socket, and documented the new json

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -216,6 +216,7 @@ list(APPEND trunk_recorder_sources
   trunk-recorder/systems/p25_parser.cc
   trunk-recorder/systems/smartnet_decode.cc
   trunk-recorder/systems/system.cc
+  trunk-recorder/recorders/recorder.cc  
   trunk-recorder/recorders/debug_recorder.cc
   trunk-recorder/recorders/analog_recorder.cc
   lib/gr_blocks/nonstop_wavfile_sink_impl.cc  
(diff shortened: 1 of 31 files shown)


robotastic:
Wow! This looks awesome!
I will have to dive into it a bit since there is a lot of change. I will try to merge this into a seperate branch so I can try it out before merging.

One thing I am not clear on is the delay creation for the file. Can you walk me through when this is helpful? Is it just so you don't have the initial empty file created when the wav-sink starts up?

amlethojalen:
There are a lot of changes, but if you go though each commit you will see that they are quite isolated.

There are no big structural changes except for the Call_Conventional, P25Convetional_Recorder, and the delayopen sink. I thought that it would be easier to follow to create separate classes and implement the changes there, otherwise there would be a whole lot of 'If conventional then' and have a bigger chance of messing up some logic.


The delay create does a few things
* dont create the file until there is activity (not empty files)
* name the file based on the start of the activity (not when it started listening)
* output a 'Listening To' when the recorder starts listening
* output a 'Recording' when the recorder detects activity
* Status socket, mark the call as 'non active' until there is activity (otherwise it was always active). 

The delay create is only implemented for the Conventional channels, it shouldn't affect any of the non conventional channels. I've only tested it on Conventional P25 channels, as that is all i'm scanning.

PR #177, opened 2018-02-11
Adds try/catch around get_mix_gain and get_lna_gain omsosdr calls

grutz:
Some drivers, such as the lime, may not support all gains and throw
errors when called. As of now only the stat_socket uploader calls
these functions.


robotastic:
Good call - I should probably add this into more sections.

PR #181, opened 2018-03-02
Debug logging some CC messages, adding unit logging

rosecitytransit:
-Changed logging of non-repetitive control channel messages from trace to debug level (based on what I've seen for P25, there may be others that are repetitive too)

-Added unit logging to radiolist.json in the capture directory

-Added commented-out code that could run an external command when a radio's state/channel changes.

Note that it would probably be best to make these configuration options,


rosecitytransit:
This was never merged so I updated it to work with the current version. Note that the config file entry is not hooked up as I am unsure how to access it in the contexts it is needed, so as is it'll always run `onunitchange.sh`.

in main.cc, the lines beginning with `sprintf(shell_command, "onunitchange.sh %li` should be something like   `sprintf(shell_command, "%s %li on &", sys->get_onunitchange_file, unit);`.

Even better, to support multiple systems, would be something like `sprintf(shell_command, "%s %s %li on &", this->get_onunitchange_file, sys->get_short_name(), unit);` along with making the config a global option.

rosecitytransit:
OK, I think I have this all ready to go.

robotastic:
Thanks - let me go give it a check tonight!

rosecitytransit:
Great! I'm wanting to set up the unit_check routine and have the option to
print out the list of radio affiliations the program keeps every so often
but the issues are a) the list gets flushed every time the program is
restarted and b) many sites have hundreds or thousands of radios and it's
much more efficient just to handle the changes.

robotastic:
The changes look good to me! Could you add in an example script that people could use to get started? You can add it to the /examples folder.

It might make the most sense to ultimately track the Unit changes in a DB. It would be easier to do the report generation and analysis over time... and no C++ to worry about!

rosecitytransit:
I got to thinking that a simple `sed` script would work, so here's a prototype/draft of one that updates both a status file (a CSV containing `radio ID,time stamp,action,talk group,acknowledgement response time stamp (if most recent action)` as well as a daily log file. Note that there may be an issue with the log file being in the recording directory before the first call of the day:
````
#!/bin/bash
printf -v TRDATE '%(%Y/%-m/%-d)T'
if $3 != "ACKRESP"
sed <capture_dir>/$1/radiolist.csv 's/^$2,.*/$2,$3,$4,/'
elif
sed <capture_dir>/$1/radiolist.csv 's/$2,.*,\d+$/'
done
echo "$2,$3,$4" >> <capture_dir>/$1/$TRDATE/radiolog.csv
````

robotastic:
Thanks! having a simple script will at least help people get started. I merged everything in and added a note in the readme.

rosecitytransit:
Just FYI, the sample script was just a draft and wasn't complete working code as-is. I can flesh it out if you don't want to. Also, I see you added a section to the read me, but you didn't delete mine (if you are wanting yours).

robotastic:
I had totally missed that in the readme!
I just switched it to your description.

If you are able to fill out the script a little, that would be great. It doesn't have to be anything fancy. I am a bad Bash scripter. What you have already written, is next level for me.

kcwebby:
> Just FYI, the sample script was just a draft and wasn't complete working code as-is. I can flesh it out if you don't want to. Also, I see you added a section to the read me, but you didn't delete mine (if you are wanting yours).

This is kind of cool -- What is the chance you would share the template for your PHP files as well?  You've shown the world the awesomeness, now we must have it :)

The bash script doesn't work, but I think I understand the intent and can fix that if I can find the time to actually sit and code.

Thanks for the commits!

rosecitytransit:
Well, thanks. The PHP script I use is at https://github.com/rosecitytransit/trunk-recorder/blob/oldtrimetscanner/unitreg.php It's not totally useful to others since it uses an API related to the radio system I record.

I'm adding an updated draft to the opened issue.

jgarvas:
@rosecitytransit Can you explain what output this might generate?  I have a custom unitlog script that I meant to submit a pull request for over a year ago and I just never could figure out how to do it.   custom TR code will generate four different events:   on, off, tg, tx.    Some of them just provide the radio ID and the event, others include the talk group.   It allows me to maintain a realtime database of every radio in the system, what talk group they're currently on, if they powered off, etc.

I'm curious if there is a way to supplement what you did with my login somehow so as not to break what you're doing, but offer the same functionality?   My code also generates these events even if the talk group in question is encrypted.  If it happens, it logs it.

rosecitytransit:
I put it in the Readme. It'll run e.g. `./script.sh shortName 123 onʼ or
`./script.sh shortName 123 call 4ʼ where 123 is a radio ID and 4 is a
talkgroup.

PR #191, opened 2018-05-16
Add osmosdr antenna support and additional json messages for call data

grutz:
1. Osmosdr support for antenna selection is added. Some devices, like the LimeSDR, have multiple options available.

2. Call output for JSON didn't include source information. This adds it back in with the downside that only the call_end record will have source data. All others will have empty fields.


dreinhold:
I get an error with my LimeMini using this :  

`Failed parsing Config: SoapyLMS7::setAntenna(TX, 1) - unknown antenna name`

robotastic:
Thanks for the submit - sorry it has taken me so long to add it in!

PR #196, opened 2018-07-13
Audio Player Code Cleanup

Dygear:
Everything works the same, it just makes it so the code is cleaner to read and is more more understandable with an general upgrade to HTML and Bootstrap.

### DIFF
--- a/audioplayer.php
+++ b/audioplayer.php
@@ -1,40 +1,63 @@
-<?php 
-$TGS = [
-    1767 => 'Example1',
-    1777 => 'Example2'
-];
+<?php
+$FileType = '.wav';
+
+$system = (empty($_GET['system'])) ? NULL : $_GET['system'];
+$date   = (isset($_GET['date'])) ? new DateTimeImmutable($_GET['date']) : new DateTimeImmutable();
+$tg     = (empty($_GET['tg'])) ? NULL : $_GET['tg'];
+
+$CONFIG = (function (string $configFilePath = './config.json') {
+    if (!file_exists($configFilePath))
+        return false;
+
+    return json_decode(file_get_contents($configFilePath));
+})();
+
+$TGFile = function (string $tgFilePath): array {
+    $return = [];
+
+    if (!file_exists($tgFilePath))
+        return $return;
+
+    foreach (file($tgFilePath) as $line) {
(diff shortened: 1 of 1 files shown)


Dygear:
1. Would `wav` and `m4a` be better?
2. Whoops! That was my bad, I didn't actually want to commit that URL.

kg6uyz:
I use MP3 with the older version of audioplayer.  I tried this new audioplayer.php but doesnt seem to work for me.  Opera/firefox cant display the page, i have to change it to a .html to get an output on the page but then the php code at the begining doesnt work.

Dygear:
@kg6uyz You have to spawn a PHP server first by running `php -S 0.0.0.0:8080` then in your browser from the same computer goto http://127.0.0.1:8080/audioplayer.php, or from another computer on the same network, simply type in the IP address of the computer it's on in place of the `127.0.0.1` address.

kg6uyz:
@Dygear Thanks, this is what i get now, 
[Mon Jul 16 07:57:12 2018] 192.168.15.126:52281 [500]: /audio.php - syntax error, unexpected '=' in /var/www/html/trunk/analog/audio.php on line 45


 Line 45 -     [$TGID, $TIME, $FREQ] = preg_split('/[-_]/', $Basename);

Dygear:
You’re gonna need at least version 7.1 to run that code.

kg6uyz:
@Dygear I fixed the issue, had to change `if (in_array($EXT, ['.m4a', '.wav']))` back to `if ($EXT != '.m4a')` and in my case i am using MP3 so i did this `if ($EXT != '.mp3')` and all is well and working.

robotastic:
Let me know when this code is working good and I can merge it in.

Dygear:
It’s working fine. There are perhaps two things that I should consider and I’d like your input on.

I’m probably going to make a define at the top of the page for the audio files to list. My default will be ‘wav’ as that’s what your program outputs by default.

The second thing is, should I make an In place configuration file reader. It’s super easy for PHP as JSON functions have been built into the core of the language since 5.2.0. That would allow for the direct read of the talk groups files as well. That I think would be the last thing to make this feature complete with the C++ code base you made.

Dygear:
@robotastic I'm happy with that one. Let me know what the rest of you think of it.

Dygear:
This one should work without any issues for everyone. Please check and advise.

robotastic:
I am merging in - feel free to add more code if you want additional changes.

PR #198, opened 2018-07-26
Update OS this runs on.

Dygear:


robotastic:
Call I will merge this in and also add in Debian! Thanks for the feedback

PR #199, opened 2018-08-01
Standardize talkgroup presentation in logs

stubbers:
Started recording messages were using the new `call->get_talkgroup_display()` for logging post #156 but alot of other messages (Not Recording, and so on) where still using `call->get_talkgroup()` resulting in inconsistent presentation of talkgroups as HEX / DECIMAL. I think I got them all


robotastic:
Good catch - Thanks!

PR #205, opened 2018-11-26
get_wacn() and get_nac() fix

frishrash:
For some reason the implementation of `get_wacn()` and `get_nac()` was flipped.


robotastic:
Good catch! Looks like the functions were only called to report stats, so shouldn't have caused errors

Issue #209, opened 2019-01-05
Feature Request: Drop extremely short calls


mdozier9:
I have six different systems that I have being monitored, and all works fairly well.  There are occasions, especially on conventional systems where something will briefly break the squelch level and trigger a very short recording (Typically less than 2 seconds).  In addition, I also see occasional key-ups on P25 systems the have a very short duration, as well as the occasional poor-performing RTL causing bad recordings.  Is it possible to create a filter to ignore these recordings, and drop/delete them?  I'm wondering if I could do a post-record upload script to look at file size and make a determination  whether to continue processing, or just delete the file.  Has anyone written a script that might do something like this?

FarvaTechnology:
We process all of our files through a MEDIA server that we build.   This media server handles the conversion of WAV to MP3 and associated tagging of the file with meta data.

If you want to ignore files less than X size - It would be simple to do it there.

You could also setup some other processes to ignore silence files all together by using various audio tools that test for the presence of sound within the file.

See also tools like: mediainfo that will tell you how many seconds a file is. (if you don't want to base it off of sound)

If you don't want to send it to a media server... you could possibly leverage the internal process of trunk-recorders JSON config file flag named: uploadScript

uploadScript gets called after every recording.  So it would be easy to test your file for presence of audio or size right there... And delete it if it's too small.

photounit:
I have been fighting the same issue for months now. I monitor all conventional frequencies and I get these ~half second bursts of noise every few minutes and it's a deal breaker for me. I had originally thought it was some type of RF bursts through the antenna but it happens simultaneously across four separate  RTL V3  dongles on VHF and UHF conventional frequencies regardless of the squelch settings which to me is too coincidental. I have tried everything to eliminate the issue from trying three separate  PC's and a laptop, eliminating my USB hub, replacing the dongles with an AirSpy and Noo Electric ones, adding a tuned helical filter on the input to 3 of the 4 dongles, and most recently putting ferrite cores on all of my USB, Ethernet, and power cables and nothing has improved my issue. 
  I agree something could probably be done in the upload script if you knew what you're doing (I don't) but it would be better if it was done before the ma4 conversion process I would think.  I was hoping maybe Luke could add a line that would give the ability to specify TR to ignore file sizes less than a certain size (or time) before the file goes through the conversion process. I had a feeling that I was not the only one experiencing this and I also suspect there are probably others too. If anyone could offer any assistance or ideas on how to fix this I would be extremely grateful.

[...]

PR #221, opened 2019-01-28
Adding the ability to set a minimum call duration, so very short calls are deleted

EricTendian:
As brought up by issue #209, there are some cases where it is useful to delete/not upload calls shorter than some duration. I've experienced this myself on an analog system where there's a lot of half-second pops that get recorded.

To address this so only real calls can be uploaded to something like OpenMHz, I have added a new config param that is set per-system, `minDuration` - this means that calls need to be this duration (in seconds) or higher to be uploaded. Below it, both the recording file and the associated call metadata JSON will be deleted.

It uses the GNURadio wav sink to get the duration of the audio file, via the recorder's `get_current_length()` method.

Also, besides implementing this new feature, I fixed a couple typos.


robotastic:
Thanks Eric - this looks good to me!

photounit:
This is awesome Eric!  THANK YOU!!!!!!

PR #229, opened 2019-02-25
Only set tia, lna and ppa gains if defined in config

kazazes:
219779f9c24ce3e585df8173f6c921ad60a4308e breaks support for devices without the new gain settings.

cc: @jareklupinski


robotastic:
Good catch

PR #236, opened 2019-05-17
Fix p25 parsing when parsing more than one p25 system with different bandplan

chuot:
Working condition: trunk recording one P25 site.

Not working condition: trunk recording more than one P25 site.

Observation while non working condition: incorrect trunk frequencies and ultimately getting "Call: more than 50 Freq." error message.

Cause: p25_parser is used as a singleton within main.cc so p25_parser::channels is getting mixed up with different bandplan frequencies from different P25 sites.

Solution: make p25_parser::channels a two dimensional array [sys_num][chan_id] then change some p25_parser methods to pass sys_num along.

### DIFF
--- a/trunk-recorder/systems/p25_parser.cc
+++ b/trunk-recorder/systems/p25_parser.cc
@@ -3,20 +3,20 @@
 
 P25Parser::P25Parser() {}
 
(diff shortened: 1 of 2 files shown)


robotastic:
Good catch!!! Thanks for finding this.

PR #242, opened 2019-06-06
Change the format of filename

vishwajan17:
Change the format of the file Name

### DIFF
--- a/trunk-recorder/call.cc
+++ b/trunk-recorder/call.cc
@@ -18,7 +18,7 @@ void Call::create_filename() {
   boost::filesystem::create_directories(path_stream.str());
 
   int nchars;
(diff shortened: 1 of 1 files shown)


robotastic:
Sorry - I am badd at my printf syntax - what does this change do?

vishwajan17:
%f prints the frequency in floating point format. %e does in scientific notation where as %g prints the shortest of above two

robotastic:
Cool - sounds like a good a change, thanks!

dotsam:
If this is being changed, all of the filenames in create_filename() should probably be changed to match as well. As the very least, the included `encode_upload.sh` sample relies on the .wav and .json file having the same filename and only different extensions, so this is now broken after this change.

PR #243, opened 2019-06-06
Change the Json file name as well

vishwajan17:
As dotsam pointed out. Fixing the json file as well to have the same name as wav file.


robotastic:
Good catch - Thanks for putting in the pull req

PR #244, opened 2019-06-10
Logging Fixes

dotsam:
A handful of logging fixes/improvements, almost entirely cosmetic, and largely bringing consistency between recorders/systems.

The biggest visual change is to also output the status of analog recorders along with digital, and prefix them in the list with "A" or "D" for analog or digital.


robotastic:
These look like good improvements, thanks!

Issue #217, opened 2019-01-12
Auto Adjust PPM


FarvaTechnology:
It would be nice to leverage the LTE-Cell-Scanner module by rxseger to automatically adjust the PPM nightly for full duty cycle trunk-recorders or provide a "run once" option to config all PPM's VS doing it manually via a GUI.

---

An ugly solution for the group is below.  Provided free by myself to the community.

Below is an ugly perl script that will leverage the rxseger LTE module to load your trunk-recorder config file.  Query LTE towers in your area and estimate your PPM for each of your RTL cards.

This could be ran nightly, weekly or just one time to get an estimated PPM for all your cards at once VS having to manually configure each card.

This script is provided FREE with no WARRANTY and it's VERY UGLY - But it works.

NOTES:
Only tested on Ubuntu - No windows version.
Only tested on RTL cards - You can tweak as needed.
YOU MUST STOP your trunk-recorder (free up your RTL cards) for this script to run.

Instructions for linux users:
1) Install  https://github.com/rxseger/LTE-Cell-Scanner

2) Make sure it works by running something like this (XX = your RTL card index number): CellSearch --freq-start 739e6 --freq-end 739e6 -iXX 

3) Once you find a freq range that works for you - Cut and past your freq values and keep them to the side.

4) cut and past the below code.

5) vi your_script_name.pl AND past this AS-IS code.

6) Modify the TOP of the program to fit your needs (specify how many RTL cards you have - and your frequency range (what you put to the side in step 3).   You can keep verbose/debug on if you want.  Up to you.

7) Make executable (or whatever perms u want) - chmod 755 your_script_name.pl

8) ./your_script_name.pl [your_json_config_file.json]  (no brackets)

9) Review output your_json_config_file.json.new

10) I would run this outside your main directory to be safe.   Tweak and share this ugly script for others to use as you wish.

Each LTE tower can take 6-30 seconds to query.  This is repeated for EACH card you have.  So for 7 RTL cards - It could take up to 2 minutes.

Depends on:
rxseger's LTE-Cell-Search
rtl_sdr (apt-get install rtl-sdr)
 

**UGLY SCRIPT**

```
#!/usr/bin/perl

### This script provided free
###
### This script is UGLY and has no warranty or claims of functionality.   USE AT YOUR OWN RISK!  I CAN'T BE RESPONSIBLE FOR ANYTHING!
###
###
### Purpose: Loads trunk-recorder like json config file.
[...]

PR #271, opened 2019-10-01
Added auto-tune functionality to recorder

kcwebby:
Luke -- I don't know how this works, but here's the pull request from the auto-ppm fork.   It's 44 commits behind and only one ahead.

### DIFF
--- a/trunk-recorder/main.cc
+++ b/trunk-recorder/main.cc
@@ -37,6 +37,7 @@
 #include <time.h>
 #include <sys/time.h>
 #include <unistd.h>
+#include <cmath>
 
 
 #include "source.h"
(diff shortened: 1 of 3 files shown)


robotastic:
Cool - I am working on getting the code cleaned up and I will try to work this in

bctrainers:
It may be worth tagging in a note that PPM doesn't work too cleanly on RTL-SDR's - and the usage of 'error' would be better.

If possible, detection via device string if sdr is in-use to have the code throw a warning back to the terminal.

kcwebby:
What issues have you had with PPM on RTLSDR?  I've been using it for some
time now?

KyleKotowick:
Hi @robotastic, just wanted to check in on the status of this. Any luck implementing some sort of auto-correct feature?

robotastic:
OK - I really need to get around to adding some form of Auto Tuning!

taclane:
I've got a branch I've been using for almost a year now to feed the FLL Band Edge outputs into a rolling average per-device.  It works great for p25 voice and control channels, but I haven't tried other systems.  Now that everything is going through the channelizer in 5.0, the method should be applicable to most systems. 

Leveraging the FLL block (which already is doing some auto-tune itself) to maintain a fine-tune correction value is likely more effective than adjusting the PPM based on message count.  The feedback from the FLL is pretty immediate, and only takes a handful of calls to bring the tuner offset under 100 Hz. 

I'll try and clean it up, and maybe we can finally close out this and #217.

robotastic:
That would be amazing!!

Dygear:
This would for sure be a nice to have. Would perhaps clear up some configuration when the device they are using isn't perfect.

taclane:
> I'll try and clean it up, and maybe we can finally close out this and #217.

Welp, Iet me see if I can do something about this before the comment is a year old.

Dygear:
With #1055 getting merged, this can probably be closed. @taclane ?

PR #287, opened 2019-12-29
Changes Formatting and Adds Color

Dygear:
I've changed the output format for some of the console logs so that the console is slighly easier to follow with the eye. I've also added color to the Recording / Not Recording text so it jumps right out at you.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 /*.json
 /*.csv
 sys_*/
+audio/
 junk/
 tester/
 fake/
--- a/trunk-recorder/call.cc
+++ b/trunk-recorder/call.cc
@@ -223,7 +223,7 @@ Recorder * Call::get_debug_recorder() {
 
 void Call::set_recorder(Recorder *r) {
   recorder = r;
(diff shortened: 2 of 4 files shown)


robotastic:
Nice upgrade - thanks

patrickfish:
Looks like this adds some blank characters before the actual talkgroup tag, example:
image

@Dygear any thoughts?

Dygear:
I wanted to make it so you could read down the information like it was a column of of data. Easier for the eye to follow.

bctrainers:
> Looks like this adds some blank characters before the actual talkgroup tag, example:
> image
> 
> @Dygear any thoughts?

Possible for you to show a screenshot without the background text bleeding in from the other terminal? From a glance of it, it would be this particular line (and section) causing the issue that you've observed: https://github.com/robotastic/trunk-recorder/commit/dbe1745546fe8fd7f09f3347e1c065a803bf04b6#diff-fa70b60f6dffff84d1b1323b90567febR489

patrickfish:
It's definitely the diff above:
https://github.com/robotastic/trunk-recorder/commit/dbe1745546fe8fd7f09f3347e1c065a803bf04b6#diff-fa70b60f6dffff84d1b1323b90567febR491

I don't see how it makes it any easier to read, but easily modified.

Issue #290, opened 2020-01-04
Airspy Gain Settings


wx-jc:
Trunk Recorder doesn't appear to support the 3 individual gain settings of the Airspy.  Maybe I'm just doing something wrong?  Instead, only "gain" works.

**Working config:**
{
        "sources": [{
                "center": 854543750,
                "rate": 10000000,
                "gain": 8,
                "digitalRecorders": 11,
                "driver": "osmosdr",
                "modulation": "qpsk",
                "device": "airspy",
                "error": 0.2

**Bad config:**

{
        "sources": [{
                "center": 854543750,
                "rate": 10000000,
                "ifGain": 8,
                "lnaGain": 8,
                "mixGain": 8,
                "digitalRecorders": 11,
                "driver": "osmosdr",
                "modulation": "qpsk",
                "device": "airspy",
                "error": 0.2

The documentation mentions support for mixGain and lnaGain.  Those don't work on their own either.  GQRX shows mix, LNA, and IF gains supported.

The regular gain setting works fine, but I think people (including me when I first set up TR) get tripped up on this with the Airspy.

n71fs:
This tripped me up as well, also I had issues using "ppm" adjustment and just ended up using "error" to do the same thing.

robotastic:
I will look through it tonight. I don't have an AirSpy, so I was just using code that someone submitted. It maybe that the way the driver handles gain has changed. Do the gain settings work in GQRX? If so, I can try to copy that code.

wx-jc:
Hi Luke,

Yes - the gain settings in GQRX work correctly.

Thanks!

Jarrod

robotastic:
Hmm... I am making the same calls as GQRX. If you check here, the values should be configurable: https://github.com/robotastic/trunk-recorder/blob/3e4ed87eac7534e4584569af3320576f02c79537/trunk-recorder/main.cc#L302
Maybe try including a `gain` setting too. If one is not included, it will default to zero which might cancel out the other settings.

wx-jc:
Weird.  

Using the regular "gain" is the only way to get it to work.  I assume it makes the other 3 settings the same value.  Luckily setting just a standard gain has been working fine for me.  It just threw me off when I first set things up a year ago.

robotastic:
I could be wrong, but I believe Gain is still a valid setting on the AirSpy. The OsmoSDR driver maybe mapping it to one of other gain settings. The other gain settings should work though and a probably worth setting on the AirSpy.

[...]

PR #293, opened 2020-01-16
Update to set the primary gain before setting secondary gains.

JoeGilkey:
This resolves issue #290 by moving the setting of the primary gain to before setting the secondary gains.


robotastic:
Thanks - let me know if this fixes things

PR #294, opened 2020-01-16
Fix for potentially unsafe sys reference.

JoeGilkey:
Moved the sys->message_count++ inside the check to see if sys is valid. Potential access of uninitialized memory otherwise.


robotastic:
Good point! Thanks for the fix

PR #298, opened 2020-02-01
Updated gitignore for Visual Studio temporary files.

JoeGilkey:
Added /.vs/ and /out/ to the gitignore file so local Visual Studio 2019 WSL files are not included.


robotastic:
These look like some interesting changes. Can you add a little on what it does and document some of the flags in the ReadMe? Also - did the MP3 file writer work? If I remember right, AAC has a little better performance at low-bitrate, but it might not be a big deal and it is a great option to have.

JoeGilkey:
Oops, the pull request should’ve been just for the VS change :( I backed out the MP3 stuff while I debug that.

JoeGilkey:
I’m cancelling this PR and will re-create it with just the actual fix.

Issue #237, opened 2019-05-21
Feature: Radio ID Tags


s3m1s0n1c:
Would be cool to add a Radio ID Tags so U can see what radio ID belongs to what unit/car etc

Eg 1234 -> Car 234

This could just populate from CSV like talkgroups.

ohaydel:
I second this feature 🥇

dbuttry:
I would love to see this feature too!

JoeGilkey:
I just put in a Pull Request that adds this to the call status output. There is a CSV file that gets loaded with the unit ID, tag and description, similar to talkgroups.

PR #299, opened 2020-02-03
MDC, FleetSync & Star signal decoding, plus Unit ID tagging.

JoeGilkey:
- Adds MDC, FleetSync & Star signal decoding.
- Adds any decoded Unit ID's to the current call source list.
- Adds emergency and tag to the specific call source.
- Adds support for Unit ID tags, similar to Talkgroup ID tags. (Issue #237)
- Updates the gitignore to support Visual Studio 2019 (easier WSL support than VS Code)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,5 @@ ChanList.csv
 cmake-build-debug/
 *.code-workspace
 /.vscode/
+/.vs/
+/out/
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -224,10 +224,17 @@ list(APPEND trunk_recorder_sources
   lib/gr_blocks/nonstop_wavfile_delayopen_sink_impl.cc
   trunk-recorder/recorders/p25conventional_recorder.cc
   trunk-recorder/recorders/p25_recorder.cc
+  trunk-recorder/csv_helper.cc
   trunk-recorder/talkgroup.cc
   trunk-recorder/talkgroups.cc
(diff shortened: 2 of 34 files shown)


bctrainers:
As for the UnitID file and it's parser...

> - **unitTagsFile** - this is a CSV files that provides information about the unit tags. It allows a Unit ID to be assigned a name. This file should be located in the same directory as the trunk-recorder executable. The format is 2 columns, the first being the decimal number of the Unit ID, the second is the Unit Name,

Has this been tested against things like TrunkPlayer where the ID persists in the JSON file, and the raw ID is imported to the PostgreSQL database? It's a neat idea and all, but not sure of the implications of other tools that hook into/utilize trunk-recorder as it currently stands.

JoeGilkey:
@bctrainers I have not checked it against TrunkPlayer. I'll check it when I get home tonight.

JoeGilkey:
@bctrainers I just looked at the TrunkRecorder source code where it handles the source list (3 separate files). This change should be compatible with how they use it. You wouldn't get the benefit of the extra fields, but I'll dig a little deeper tonight and maybe I can add a pull request for them to take advantage of the fields. At the very least though, this change will not break their system.

bctrainers:
Good to hear. :)

JoeGilkey:
@bctrainers I was able to verify that this change doesn’t interfere with TrunkPlayer. I also checked it against my next pull request, which is Motorola TPS signaling (what FDNY uses), but that change will wait until this one is in.

PR #301, opened 2020-02-04
Fdkaac

robotastic:
Switching to FDKAAC instead of ffmpeg

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,12 @@
 Trunk Recorder ChangeLog
 ========================
 
+### Version 3.1.1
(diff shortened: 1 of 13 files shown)


robotastic:
@gofaster Yea... that probably is a bug. I was trying to have 2 Call objects at the same time, one for each freq the Call was active on. I will have to go double check the logic.

robotastic:
@gofaster Dug into it a bit more and it looks like what is getting printed is actually wrong and the code was doing the right thing. Only a single additional recorder gets started... I re-arrange the print-outs to make this clearer.

PR #303, opened 2020-02-04
Motorola TPS Signaling, and some cleanup.

JoeGilkey:
This PR adds the elusive Motorola TPS signaling, used by FDNY. It is a P25 frame with the Unit ID, with analog audio. To accomplish this, I have refactored some of the earlier decoder code I had added, and created a new tps decoder in gr_blocks/decoders. Here's a list of what's changed:

- Added tps_decoder_sink and related files.
- Moved decoders into a /decoders/ directory under gr_blocks.
- Added decodeTPS option in system configuration.
- Added decoder_wrapper and related files to handle the different types of decoders (gnuradio based and non-gnuradio based)
- Added logLevel option in configuration, and support for changing the logging level (for debugging purposes).
- Added system_type to the Call_Source entries in cases where the entry doesn't come from a signaling decoder.
- Added it_tcp_sink to allow a Source in trunk-recorder to look like a rtl_tcp device. It does not allow control of the Source (yet) but it is useful for setting up on headless configurations (ppm/error/gain settings, etc.)


robotastic:
Cool - I will go give this a try tonight and look to merge it in. Thanks for all code you have been writing!

JoeGilkey:
@robotastic thanks! :)

robotastic:
I like the idea of being able to view what a source is doing! I am not sure if it is working right, though. I have attached 2 screenshots. The first is of GQRX capturing the spectrum directly from an RTL. The second is using TR as the source. When it is getting the IQ from TR, it looks like something is off. Could there be an issue with variable size or type?
Screen Shot 2020-02-05 at 10 05 41 PM
Screen Shot 2020-02-05 at 10 07 53 PM

To make things easy, what if we use [UDP Sink](https://www.gnuradio.org/doc/doxygen/classgr_1_1blocks_1_1udp__sink.html) and sent the output to a small Python program we could build using GnuradioCompanion? 
This way we could sample further down the flow graph. This is sort of what the Debug Recorder: https://github.com/robotastic/trunk-recorder/blob/master/trunk-recorder/recorders/debug_recorder.cc and the Debug graph do: https://github.com/robotastic/trunk-recorder/blob/master/debug.grc

Instead of saving a file, we could send the channel down to be graphed. That would help alot for adjusting error and gain. You can probably get a similar experience with GQRX by zooming in on the FFT and correctly centering.

JoeGilkey:
I think your right, the sizing of the data elements is off. I’ll play around with it over the weekend.

robotastic:
Cool - I will also try resturcturing the debug-recorder so you can see what is happening at the channel level.

JoeGilkey:
Ok, I figured it out. Sorry I missed this, but it was very simple. Gr-OsmoSDR converts the raw i/q bytes into a gr_complex. For iq_tc_sink to work, it needs to convert back from gr_complex to i/q bytes. I do like the idea of using a udp sink, I'll look into that for a future pull request. :)

JoeGilkey:
I took out the TCP Sink, I'll do a separate pull request for that.

PR #304, opened 2020-02-04
Set final length in call status JSON too

dotsam:
Busy time around here with pull requests! Thought I'd throw this one in too to set the final call length in the status json file as well, for those of us not using openmhz


JoeGilkey:
Should probably change the  final_length to this->final_length to keep it consistent. I was also thinking that since there's several places where the call info is output, it may be helpful to have a single method that they all call that writes the call to a stream as a full json block.

robotastic:
That makes sense - let me go see if I can put together a more unified way to generate all this.

robotastic:
Finally added this in with commit: https://github.com/robotastic/trunk-recorder/commit/dc90ae4ff98504b58cb15e0d053926016a215d2b

Sorry this took soooooooooooooooo long for me to do!

Issue #318, opened 2020-02-15
P25Conventional stops recording


gofaster:
Monitoring a conventional P25 system with delayCreateOutput enabled.
A P25 recorder listens for a call, then starts a recorder and stops it afterwards.  This works fine for a while.
This system transmits it's call sign as Morse code via plain FM every 30 mins.  
All P25 calls after this station ID transmission will not trigger a P25 recorder to start.  It stays in the "Listening P25 recorder" state and regularly increments the elapsed time counter at the usual 200 sec interval.
This situation is immediately corrected by restarting the trunk-recorder instance - and works normally until after the next station ID transmission.  
Something stuck in the flowgraph and blocking?

gofaster:
Tested with delayCreateOutput set to false.  Same issue, all P25 calls after the station ID transmission are not recorded.
![station id](https://user-images.githubusercontent.com/1273372/74608353-4b88a780-50ae-11ea-9d28-926277a4901b.png)

gofaster:
With P25 verbosity set at 10 - before a station ID transmission, P25 calls emit lines like the following.  All calls are recorded and voice decodes sound great.
There are no lines emitted when the station ID is transmitted.
All subsequent P25 calls do not emit any lines.
Is it possible to add debug logging statements to the blocks in this part of the flowgraph to see where it dies?

```
...
1581888801.760599 NAC 0x114 TDU3:
1581888801.760609 NAC 0x114 TDU3:
1581888801.760669 NAC 0x114 TDU3:
1581888801.818591 NAC 0x114 TDU3:
1581888801.818635 NAC 0x114 TDU3:
1581888801.818645 NAC 0x114 TDU3:
1581888801.818654 NAC 0x114 TDU3:
1581888801.818664 NAC 0x114 TDU3:
1581888801.915752 NAC 0x114 TDU3:
1581888801.915795 NAC 0x114 TDU3:
1581888801.915806 NAC 0x114 TDU3:
1581888801.915815 NAC 0x114 TDU3:
1581888801.915824 NAC 0x114 TDU3:
1581888801.973320 NAC 0x114 TDU3:
1581888801.973367 NAC 0x114 TDU3:
1581888801.973377 NAC 0x114 TDU3:
1581888801.973386 NAC 0x114 TDU3:
1581888801.973414 NAC 0x114 TDU3:
1581888802.072238 NAC 0x114 TDU3:
1581888802.072278 NAC 0x114 TDU3:
[...]

PR #322, opened 2020-02-29
Workaround for P25Conventional recorder going deaf & update analog/P25Conv call start times.

gofaster:
Workaround for [https://github.com/robotastic/trunk-recorder/issues/318](url) resets the costas_clock  on a periodic basis (currently 200 sec) in case it has gone into an errored state.    This only applies to P25Conventional recorders with QPSK modulation if they are idle.

Add system number prefix to log lines.

Add Talkgroup number and priority to log for calls that are rejected.


gofaster:
For analog and P25Conventional recorders, changes to make reported call start times more reflective of reality vs file creation time.   The updated call start time is stored in the JSON metadata file.  The original wav and JSON filenames (with the encoded time) are not renamed with the updated time in case existing scripts and downstream apps rely on them.   

TRUNK-PLAYER note:  At this time, though trunk-player appears to read the start_time from the JSON file, it still stores or displays the call start time based on the timestamp in the filename.  Some massaging in the upload script is required to rename the wav and JSON file with the updated timestamp before passing it to manage.py

robotastic:
Nice work - I will go give this a look through and try. It makes sense to me, I will just see if there are some ways to streamline things.

gofaster:
src_list entries from a signalling decoder (MDC1200) were not cleared when an analog_recorder restarts.  It keeps growing with each consecutive call.
Added clear function.

gofaster:
Add change to enable decoding of Unit ID's for Conventional P25 calls.

gofaster:
@EricTendian This PR got a bit unwieldy as I kept making  changes after the initial PR.  Will try to resolve the merge conflicts with the latest upstream master.

EricTendian:
@gofaster just wondering if you had a chance to do that yet - I'm happy to help since I've already been running the changes on a custom build of mine.

gofaster:
@EricTendian Sorry,  my time and attention has been needed elsewhere recently.  I had wanted to incorporate the changes into the gnuradio 3.8 branch.  If you feel like implementing the changes, I would appreciate it.

robotastic:
Would it be possible to just reset the Costas Clock at the end and/or beginning of each call?

I just made this change to reset it at the start of each QPSK  P25 Conventional Call. Would that have the same effect?

EricTendian:
@gofaster no problem at all! I figured you had been busy. 

I made a PR for the unit ID related changes: https://github.com/robotastic/trunk-recorder/pull/358. @robotastic I think you introduced a compile error with the costas_clock change, I fixed it for you in that PR.

Will be making another for the improved call start time after that one is merged. Will wait for the question above to be answered before doing anything with the Costas Clock.

gofaster:
@EricTendian   Thank you!  Correcting the analog call start time made matching up (multi-agency) incident responses much less fuzzy.

@robotastic  If I recall, that won't fix it.  When the Costas Clock gets stuck in it's rotated state, the downstream P25 Frame Assembler block stalls forever looking for a valid P25 Frame Sync magic pattern; it will never produce any output to the downstream blocks.   Even if a legit p25 call is subsequently received, the output symbol stream is nonsensical  until Costas Clock is reset.
As far as TR is concerned, the call is never set active or reaches the is_idle() check to stop and (re)start the recorder.
The proper fix would be in the Costas block itself.  Other places would break the modularity of the flowgraph/TR  Perhaps a separate watchdog thread?

PR #324, opened 2020-03-04
Silence fdkaac output to avoid unnecessary log messages when uploading calls to OpenMHz

EricTendian:
When a call is uploaded to OpenMHz after the switch was made to fdkaac, the following log messages occur amongst other logs:

```
[100%] 00:14.290/00:14.290 (246x), ETA 00:00.000   
114322/114322 samples processed in 00:00.058
```

To clean up the output, I'm adding the `--silent` argument, after which these messages aren't shown.


robotastic:
Good call!

PR #329, opened 2020-04-20
Adding built-in support for uploading to Broadcastify Calls

### LINKED ISSUE #327: Feature request: Upload to Broadcastify Calls
See: https://wiki.radioreference.com/index.php/Broadcastify-Calls-API

An integrated option to upload to the Broadcastify Calls Platform.  

https://www.broadcastify.com/calls

EricTendian:
This PR introduces a way to upload calls to Broadcastify Calls without needing extra scripts. A user would update their `config.json` with new parameters mentioned in the README file for uploading to work. This has been developed to work alongside of OpenMHz.com uploads, instead of as a replacement to it.

I have this running on a system as a demonstration: https://www.broadcastify.com/calls-dev/systems/7

To handle the uploads, I added a dependency on `libcurl` since that seems to be much safer than using a custom HTTP implementation as in `uploader.cc`. Ideally I would refactor the OpenMHz logic to use libcurl as well, but I first wanted to see if this would be accepted.

More specific details on changes will be provided in a self-review of the code. This is a draft PR since I need to do a bit more testing on how the new HTTP logic performs under different conditions (disconnected network, server outage, etc).

Fixes #327.


trunk-recorder/uploaders/call_uploader.cc
```diff
@@ -119,6 +120,247 @@ void build_call_request(struct call_data_t *call, boost::asio::streambuf& reques
...
+    curl_easy_cleanup(curl);
+
+    /* then cleanup the formpost chain */
+    curl_formfree(formpost);
+
+    /* free slist */
+    curl_slist_free_all(headerlist);
+
+    if (res == CURLM_OK) {
+      std::size_t spacepos = response_buffer.find(' ');
```
EricTendian, line 344: If we get a successful response according to curl, we're checking the string to try and parse out the response from Broadcastify in order to upload the actual audio file.

trunk-recorder/uploaders/call_uploader.cc
```diff
@@ -129,7 +371,7 @@ void convert_upload_call(call_data_t *call_info, server_data_t *server_info) {
   int nchars = snprintf(shell_command, 400, "sox %s -t wav - --norm=-3 | fdkaac --ignorelength -b 8000 -o %s -", call_info->filename, call_info->converted);
 
   if (nchars >= 400) {
-    BOOST_LOG_TRIVIAL(error) << "Call Uploader: Path longer than 400 charecters";
+    BOOST_LOG_TRIVIAL(error) << "Call Uploader: Path longer than 400 characters";
```
EricTendian, line 374: Fixed typo

trunk-recorder/uploaders/call_uploader.cc
```diff
@@ -139,6 +381,18 @@ void convert_upload_call(call_data_t *call_info, server_data_t *server_info) {
 
   // BOOST_LOG_TRIVIAL(info) << "Finished converting\n";
 
+  if (call_info->bcfy_calls_server != "" && call_info->bcfy_api_key != "" && call_info->bcfy_system_id > 0) {
+    upload_to_broadcastify(call_info);
+  }
+
+  if (call_info->upload_server == "") {
```
EricTendian, line 388: This is needed in case we only wanted to upload to Broadcastify.

trunk-recorder/uploaders/call_uploader.cc
```diff
@@ -147,23 +401,24 @@ void convert_upload_call(call_data_t *call_info, server_data_t *server_info) {
 
   size_t req_size = request_.size();
 
+  int error;
+
   if (call_info->scheme == "http") {
-    BOOST_LOG_TRIVIAL(info) <<"[" << call_info->short_name <<  "]\tTG: " << call_info->talkgroup << "\tFreq: " << FormatFreq(call_info->freq)  << "\tHTTP Upload result: " << http_upload(server_info, request_);
```
EricTendian, line 151: This didn't seem very DRY so refactored this to use the same logging as HTTPS.

trunk-recorder/uploaders/call_uploader.cc
```diff
@@ -207,23 +462,26 @@ void send_call(Call *call, System *sys, Config config) {
   boost::regex  ex("(http|https)://([^/ :]+):?([^/ ]*)(/?[^ #?]*)\\x3f?([^ #]*)#?([^ ]*)");
   boost::cmatch what;
 
-  if (regex_match(config.upload_server.c_str(), what, ex))
-  {
+  strcpy(call_info->filename,  call->get_filename());
```
EricTendian, line 465: Moved this up so we'd still set these fields if we're only uploading to Broadcastify.

CMakeLists.txt
```diff
@@ -103,6 +103,7 @@ find_package(GrOsmoSDR)
 find_package(LibHackRF)
 find_package(UHD)
 find_package(OpenSSL REQUIRED)
+find_package(CURL REQUIRED)
```
EricTendian, line 106: I figure if curl is eventually going to replace the custom HTTP logic, then this should be required. The parent Docker image should be updated though to include the dev headers for curl.

trunk-recorder/call.cc
```diff
@@ -121,6 +121,8 @@ void Call::end_call() {
     BOOST_LOG_TRIVIAL(info) << "[" << sys->get_short_name() << "]\tTG: " << this->get_talkgroup_display() << "\tFreq: " << FormatFreq(get_freq()) << "\tEnding Recorded Call - Last Update: " << this->since_last_update() << "s\tCall Elapsed: " << this->elapsed();
 
     final_length = recorder->get_current_length();
+    // Fixes https://github.com/robotastic/trunk-recorder/issues/103#issuecomment-284825841
+    start_time = stop_time - final_length;
```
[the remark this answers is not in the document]
robotastic: Interesting - let me dig into this. The Start time on trunking systems should be when the call goes on the control channel. Since the `final_length` excludes some of the dead area time, this way of calculating start_time maybe off.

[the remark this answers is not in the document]
EricTendian: Ah, this might be something we only want to do on conventional channels then, with long-running recorders.

robotastic:
Cool! I am not against a switch to CURL. The Boost HTTP Libraries are sort of a mess. Let me do a little research on libCurl, I have really only used the command line stuff.

bctrainers:
As seeing we're offering more than one place to send json and audio data to now, it may be worthwhile placing OpenMHZ and Broadcastify code into their own dedicated files/directories in a modular fashion.  With such, we could then use something to exclude or include said modules in the build process on the terminal when compiling. The same idea could be applied to which type of radio systems we want to monitor.

```bash
--remote-api=openmhz,broadcastify 
--system-types=p25,smart,analog```

However, not sure how reliable such operators would be currently.

patrickfish:
I think leaving them as config options is the way to go. There’s no huge
benefit to not have it compiled at build time, and then as you want to
adjust your configs in the future you don’t need to recompile. The build
time for these modules is negligible.

trunk-recorder/call.cc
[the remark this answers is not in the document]
gofaster: I had done some changes to account for idle (dead air) time between calls on conventional and P25conventional calls
[https://github.com/gofaster/trunk-recorder/commit/1c502b3b5bca0f0ba365864ea3e1492fba67a048](url)
[https://github.com/gofaster/trunk-recorder/commit/e2122a0d2ae907ac0b19deef6567bd930bf6dd08](url)
[https://github.com/gofaster/trunk-recorder/commit/a3a8ba5635b8c89dc4b3ab9861d48280a04e1eb6](url)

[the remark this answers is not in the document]
EricTendian: @gofaster thanks, that seems like the right approach. Your solution seems more robust so I'd prefer #322 to be merged and I can pull my more naive fix out.

EricTendian:
@robotastic @bctrainers good ideas. What I'll do is refactor the uploader logic so there's a class for an OpenMHz uploader, and a class for a Broadcastify uploader, both using libcurl for the underlying API calls. (config logic remains as-is)

I think ideally the uploading logic could be in the style of a webhook, a common API call format that could be ingested by various player servers. Then for anything that needs a more custom implementation, the existing shell script approach can work.

robotastic:
I did a quick scan through and things generally look pretty sensible. libcurl also seems to have great reviews all around. I will try to get things compiled tomorrow night to give it a test run.

EricTendian:
Update on my above comment:

I was able to refactor the OpenMHz logic to use libcurl as well, now both Broadcastify and OpenMHz implement a parent Uploader class. Future uploaders can be added by adding the appropriate config functions as well as a new child Uploader class.

I also got a chance to test out different error cases, I think I have the common ones handled.

Since there was some discussion about the call length change, I removed it from this PR so it can be introduced separately.

Here's two systems running this branch's build in Docker:

- Broadcastify
  - https://www.broadcastify.com/calls-dev/systems/7
  - https://www.broadcastify.com/calls-dev/systems/8
- OpenMHz
  - https://openmhz.com/system/sc21102
  - https://openmhz.com/system/chi_cfd

EricTendian:
@robotastic just wanted to check in and see if you had a chance to test this out?

By the way, this branch is now in use for uploading to Broadcastify's production calls site, https://www.broadcastify.com/calls/status/ - I stream Chicago Fire, Chicago Police, and STARCOM21. (Still uploading to OpenMHz as well)

robotastic:
Awesome to hear it is running. I am going to check it out this weekend. I looked through the code and it made sense to me, I want to double check it compile and works.

robotastic:
Thanks @EricTendian - I think it is running well and the code changes look good. I read through the code and it generally makes sense to me. Let's see how it goes!

EricTendian:
@robotastic great, thank you!

robotastic:
@EricTendian I am actually having a problem compiling now on my Mac. Could you give a try compiling Master? It seems to work fine on Linux, but not OSX. Any idea what `basename()` function was?

This is the error:
````
[ 65%] Building CXX object CMakeFiles/recorder.dir/trunk-recorder/uploaders/broadcastify_uploader.cc.o
/Users/luke/Programming/trunk-recorder/trunk-recorder/uploaders/broadcastify_uploader.cc:90:39: error: use of undeclared identifier 'basename'; did you mean
      'boost::filesystem::basename'?
               CURLFORM_COPYCONTENTS, basename(call->converted),
                                      ^~~~~~~~
                                      boost::filesystem::basename
/opt/local/include/boost/filesystem/convenience.hpp:39:24: note: 'boost::filesystem::basename' declared here
    inline std::string basename(const path & p)
                       ^
/Users/luke/Programming/trunk-recorder/trunk-recorder/uploaders/broadcastify_uploader.cc:90:39: error: cannot pass object of non-trivial type 'std::string' (aka
      'basic_string<char, char_traits<char>, allocator<char> >') through variadic function; call will abort at runtime [-Wnon-pod-varargs]
               CURLFORM_COPYCONTENTS, basename(call->converted),
                                      ^
2 errors generated.
make[2]: *** [CMakeFiles/recorder.dir/trunk-recorder/uploaders/broadcastify_uploader.cc.o] Error 1
make[1]: *** [CMakeFiles/recorder.dir/all] Error 2
make: *** [all] Error 2
````

EricTendian:
@robotastic ah, that's my fault. Looks like I happened to pull that from Linux's standard library. The boost suggestion you got seems to need a different datatype so not a 1:1 match. I removed it in https://github.com/robotastic/trunk-recorder/pull/332 since it wasn't necessary, tested both Broadcastify and OpenMHz without it and it worked fine.

robotastic:
Perfect! that fixed it. Thanks for the quick fix!

PR #330, opened 2020-04-22
Cleanup signaling with callbacks

JoeGilkey:
The current code uses a reference to the current call in the signaling decoders. This isn't exactly "clean" since the decoders are in the "libs" directory. To fix this, I've switched the code to using callbacks instead. In addition, this code performs the following:
* Fixes the build warnings (unsafe variables, etc).
* Adds the ability to broadcast signals thru the status socket as a "signaling" message type.
* Adds a configuration setting to turn the "signaling" messages on and off.
* Updates the documentation to add the "signaling" message.
* Updates signaling to include the type of message received, not just the emergency true/false. These types are generic and supported by most signaling systems.
* Adds an on_message handler to stats_socket so the acknowledgements don't build up from TrunkPlayer.


robotastic:
Nice work - I am going to try and get all the new PRs in this weekend. I don't have a great way of testing the signaling, but I will make sure things run normally.

JoeGilkey:
@robotastic I've updated this PR to resolve the merge conflict in the Readme.md file.

robotastic:
Thanks! looks good to me. I can't test if the signaling works, but let me know if there are any issues.

PR #332, opened 2020-04-26
Removing usages of basename for MacOS compatibility

EricTendian:
Turns out that `basename()` as used is coming from the Linux standard library, which obviously isn't available in MacOS. It wasn't necessary (just a potentially helpful way to obfuscate the true path of the file), so I'm removing it here.


robotastic:
Awesome! Thanks!

PR #348, opened 2020-05-03
Making callLog optional for Broadcastify Calls to work, better error handling for missing call upload deps

EricTendian:
This PR adds two enhancements to provide a smoother experience when using trunk-recorder only for Broadcastify Calls.

First, if a user forgets to install sox or fdkaac, and so the wav conversion fails, the upload thread will alert the user and not continue on to try and upload the call. Here's what this looks like:

```
[2020-05-03 18:17:31.857895] (info)   	- Starting P25 Recorder Num [0]	TG:       9008	Freq: 770.943750 MHz 	TDMA: false	Slot: 0
[2020-05-03 18:17:31.858128] (info)   [sc21102]	TG:       9008	Freq: 770.943750 MHz	Starting Recorder on Src: rtl=1,buflen=65536
[2020-05-03 18:17:32.750410] (info)   [sc21102]	TG:       9008	Freq: 772.643750 MHz	Update Retuning - New Freq: 772.643750 MHz	Elapsed: 1s 	Since update: 0s
[2020-05-03 18:17:33.003577] (info)   [sc21102]	TG:       3981	Freq: 773.531250 MHz	Ending Recorded Call - Last Update: 4s	Call Elapsed: A
[2020-05-03 18:17:33.003776] (info)   	- Stopping P25 Recorder Num [1]	TG:       3981	Freq: 773.531250 MHz 	TDMA: false	Slot: 0
sh: 1: sox: not found
ERROR: unsupported input file
[2020-05-03 18:17:33.012433] (error)   Failed to convert call recording, see above error. Make sure you have sox and fdkaac installed.
[2020-05-03 18:17:37.004152] (info)   [sc21102]	TG:       9008	Freq: 772.643750 MHz	Ending Recorded Call - Last Update: 4s	Call Elapsed: 6
[2020-05-03 18:17:37.004470] (info)   	- Stopping P25 Recorder Num [0]	TG:       9008	Freq: 772.643750 MHz 	TDMA: false	Slot: 0
sh: 1: sox: not found
ERROR: unsupported input file
[2020-05-03 18:17:37.012710] (error)   Failed to convert call recording, see above error. Make sure you have sox and fdkaac installed.
[2020-05-03 18:17:57.601939] (info)   [sc21102]	TG:       2753	Freq: 773.293750 MHz	TG not in Talkgroup File 
[2020-05-03 18:17:57.602028] (info)   	- Starting P25 Recorder Num [0]	TG:       2753	Freq: 773.293750 MHz 	TDMA: false	Slot: 0
[2020-05-03 18:17:57.602252] (info)   [sc21102]	TG:       2753	Freq: 773.293750 MHz	Starting Recorder on Src: rtl=1,buflen=65536
```

Second, it changes the behavior of the `callLog` config option slightly, so that a JSON file is always generated, but like `audioArchive`, when `callLog` is false the JSON will get deleted after the builtin call upload as well as any other upload scripts run. This means that both audioArchive and callLog can be false, and uploads will still work.

I tested this by turning both off and doing `watch ls -lt` within the appropriate media directory, seeing that the only new files created were temporary ones for the current call being recorded/uploaded.

The diff looks a bit bigger than it actually is because there's whitespace changes from unindenting a block.

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,7 +121,7 @@ Here are the different arguments:
    - **broadcastifyApiKey** - *(Optional)* System-specific API key for Broadcastify Calls
    - **broadcastifySystemId** - *(Optional)* System ID for Broadcastify Calls (this is an integer, and different from the RadioReference system ID)
    - **audioArchive** - should the recorded audio files be kept after successfully uploading them. The options are *true* and *false* (without quotes). The default is *true*.
-   - **callLog** - should a json file with the call details be generated. The options are *true* and *false* (without quotes). The default is *true*. This is required if uploading to Broadcastify Calls or trunk-player.
+   - **callLog** - should a json file with the call details be kept after successful uploads. The options are *true* and *false* (without quotes). The default is *true*.
    - **minDuration** - the minimum call (transmission) duration in seconds (decimals allowed), calls below this number will have recordings deleted and will not be uploaded. The default is *0* (no minimum duration).
    - **bandplan** - [SmartNet only] this is the SmartNet bandplan that will be used. The options are *800_standard*, *800_reband*, *800_splinter*, and *400_custom*. *800_standard* is the default.
    - **bandplanBase** - [SmartNet, 400_custom only] this is for the *400_custom* bandplan only. This is the base frequency, specified in Hz.
--- a/trunk-recorder/call.cc
+++ b/trunk-recorder/call.cc
@@ -131,40 +131,37 @@ void Call::end_call() {
       freq_list[freq_count - 1].error_count = rx_status.error_count;
     }
 
-    if (sys->get_call_log()) {
-      std::ofstream myfile(status_filename);
+    std::ofstream myfile(status_filename);
 
 
-      if (myfile.is_open())
-      {
-        myfile << "{\n";
-        myfile << "\"freq\": " << this->curr_freq << ",\n";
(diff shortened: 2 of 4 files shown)


robotastic:
So I was thinking of adding a archiveCallLog flag. I have my system on an SSD and it would be better not to do tons of writes/deletes if it is not needed. The same is true for systems running on MicroSD cards, like the Pis. I do like the idea of having giving warnings is SOX is not present. Do you think it should just print a warning or quit?

EricTendian:
@robotastic ah, good points. I was thinking long-term the best option might be to keep the call data (JSON, and maybe audio too) in memory, only writing them out to disk if needed - so any upload logic would just read the JSON string in memory instead of from the filesystem.

As for warning vs quit, I can see an argument for either way - I went with a warning since it seems to be how other failures are handled currently, but if we were to detect if those binaries were missing during config parsing that could also work. Related to the above comment, if we end up doing more of this in memory rather than in the filesystem, I can see a situation where we use `libsox` and `libfdk-aac` rather than the command line tools themselves, and in that case we would check for those dependencies at compile-time.

Right now my changes are a more short-term attempt to fix the issue, since the above ideas would take more refactoring.

robotastic:
Cool - this looks good to me, sorry for the delay in merging. I wonder if there is some way to just pipe the JSON right into the HTTP request if a file isn't going to be saved.

PR #350, opened 2020-05-04
Updating README to indicate IF gain also applies to Airspy

EricTendian:
I noticed in the README that there were only two gain modes for the Airspy, however it has three documented ones elsewhere - IF, MIX, and LNA. Thus, I'm making this update to the README to indicate that `ifGain` also works for the Airspy.

I tested this on my own Airspy and also verified it in the gr-osmosdr source: https://github.com/osmocom/gr-osmosdr/blob/v0.1.4/lib/airspy/airspy_source_c.cc#L505

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,7 +91,7 @@ Here are the different arguments:
    - **squelch** - Analog Squelch, my rtl-sdr's are around -60. [0 = Disabled] _Squelch needs to be set if the System using the source is conventional._
    - **error** - the tuning error for the SDR in Hz. This is the difference between the target value and the actual value. So if you wanted to recv 856MHz but you had to tune your SDR to 855MHz to actually receive it, you would set this to -1000000. You should also probably get a new SDR if it is off by this much.
    - **gain** - the RF gain to set the SDR to. Use a program like GQRX to find a good value.
-   - **ifGain** - [hackrf only] sets the if gain.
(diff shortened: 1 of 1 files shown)


bctrainers:
@EricTendian, are you on the gitter.im chat perhaps? If you are, i couldn't find your username. :)

It might be best to consolidate most of the README.md stuff into the wiki?

EricTendian:
@bctrainers I like that idea, however is there a way to tie a wiki update to a pull request? In the past I've put all my documentation in a README because in cases like this I can update documentation in the same pull request as adding a new feature, not needing to remember to update the wiki after my pull request gets merged.

bctrainers:
Not that I am aware of right now, the wiki is open-access editing to all. 

The wiki does have a git repo of its own, but not something you're looking for if i am not mistaken. 

https://help.github.com/en/github/building-a-strong-community/adding-or-editing-wiki-pages#adding-or-editing-wiki-pages-locally

robotastic:
Ah! Good call - I forgot to add this back into the readme. Good to hear it works, I didn't have an Airspy to test on. I do like to keep the arguments in the readme, since in theory it stays align with the current version of the software. It makes it a little easier to go back through the code and see when a feature was added. For everything else, the wiki seem to be a good option. I just have to remember to add to it!

PR #351, opened 2020-05-04
Adding auto gain control option to source if it supports it

EricTendian:
Since some SDRs (like the RTL-SDR) support the ability for AGC to be turned on/off, I figured that setting should be available in the config. Added it to the relevant spots, as well as updated the README. Tested it on my own RTL-SDR.

I included a warning on it since I've read that most SDR AGC's don't work that well for the kind of narrowband monitoring done here.

I also added a line for ppm in the README since people have already been using it but it wasn't officially documented.


robotastic:
I am a little nervous to add AGC because as you noted, it tends to create more problems than it solves. I suppose it can't hurt to try - and we can always remove it if people are using it unwisely.

Issue #311, opened 2020-02-08
Issue with recording P25 Phase 2 traffic


CqDx914:
Hello, 

I have been using a Raspberry Pi 3 B+ with trunk-recorder installed and it’s doing a great job on decoding and recording P25 Phase 1 traffic, but when the network recently started carrying Phase 2 traffic, all it records were garble and unintelligible traffic. I see the logs reflect it recognizes TDMA traffic but it just would not record Phase 2 intelligibly.

Anyone has experienced this as well?

My raspberry pi is currently running on the following
Linux version 4.14.79-v7+ (dc4@dc4-XPS13-9333) (gcc version 4.9.3 (crosstool-NG crosstool-ng-1.22.0-88-g8460611)) #1159 SMP Sun Nov 4 17:50:20 GMT 2018

Here is a sample of the trunk-recorder log that I captured:
[2020-02-07 21:57:46.454713] (info)   [METRO25]	TG: 9032	Freq: 8.530375e+08	TG not in Talkgroup File 
[2020-02-07 21:57:46.455067] (info)   	 P25 Recorder - Adjust Channel Resampled Rate: 30117.6 System Rate: 30000 ARB Rate: 0.996094
[2020-02-07 21:57:46.455979] (info)   	- Starting P25 Recorder Num [0]	TG: 9032	Freq: 8.530375e+08 	TDMA: true	Slot: 0
[2020-02-07 21:57:46.457004] (info)   [METRO25]	TG: 9032	Freq: 8.530375e+08	Starting Recorder on Src: 
[2020-02-07 21:57:52.001970] (info)   [METRO25]	TG: 9032	Freq: 8.530375e+08	Ending Recorded Call - Last Update: 4s	Call Elapsed: 6
[2020-02-07 21:57:52.015310] (info)   	- Stopping P25 Recorder Num [0]	TG: 9032	Freq: 8.530375e+08 	TDMA: true	Slot: 0

robotastic:
It is definitely possible that the math gets mixed up on ARM chips. The only real difference between OP25 and Trunk Recorder for Phase 2, are these files: https://github.com/robotastic/trunk-recorder/tree/master/lib/lfsr
I will try to look through and see if I spot any obvious things where it is assuming

GTR8000:
I also think that Phase II decoding is hosed up on the ARM platform.

[...]

PR #355, opened 2020-05-11
Fix for Issue #311

jonk6:
Resolves problem of garbled P25 Phase2 TDMA messages.
    
    There appear to be integer rollovers that occur in the Raspberry Pi environment,
    but not in the X86 environment because:
    
        x86:  __SIZEOF_LONG__ = 8
        rpi4: __SIZEOF_LONG__ = 4
    
    Modifications to lib/lfsr/bit_utils.h, lib/lfsr/lfsr.cxx, and lib/lfsr/lfsr.h
    to use unsigned long long instead of unsigned long resolves the problem for the
    RPi world.
    
    Since
    
        x86:  __SIZEOF_LONG_LONG__ = 8
        rpi4: __SIZEOF_LONG_LONG__ = 8
    
    it doesn't break anything for X86.

### DIFF
--- a/lib/lfsr/bit_utils.h
+++ b/lib/lfsr/bit_utils.h
@@ -53,7 +53,7 @@ Eigen::VectorXi* dibits_to_bits(Eigen::VectorXi &dibits) {
 }
 
 // Tested OK
-Eigen::VectorXi *mk_array(unsigned long n, unsigned l) {
+Eigen::VectorXi *mk_array(unsigned long long n, unsigned l) {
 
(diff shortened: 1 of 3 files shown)


robotastic:
Great catch! This makes a lot of sense. I am going to roll this in. I don't have good systems to test on, so if anyone has TDMA issues from this merge, let me know.

Dygear:
I'm about to check and advise on this. (Commenting so I know where to remote back to.)

PR #357, opened 2020-05-19
Adding maximum call duration to split long calls

### LINKED ISSUE #197: Enhancement: Implement a Mechanism to End Recording If A User-Defined Length is Exceeded
### The Problem
Certain trunked systems are configured so that when a talk group is in emergency mode the assigned repeater will be held active and a new channel grant is not issued when a unit keys up. That is, all activity continues on one frequency until the emergency status is released. This can make for lengthy recordings, and make near real-time monitoring impossible.

### Proposed Solution
Allow the user to specify a maximum recording time in seconds. Once exceeded, begin looking for a good place to stop recording and start another file. A "logical break" could be inserted when a Termination Data Unit (TDU) is detected on the voice channel. If I understand correctly, a TDU follows the comletion of every voice message (See page 53 of http://www.dvsinc.com/papers/p25_training_guide.pdf)

### Assumptions
Trunk Recorder is already monitoring the low speed data signal on the voice channel. This is assumed because time indicators of units beginning transmissions are provided in the recording's json file.

### Anticipated changes
1. Recording code would need to be updated to check for recording length and whether it specifies the user-defined parameter; code added to terminate the recording and start a new file in the same way as happens when the call timeout is exceeded.
2. A parameter added to the system record in config.json, maximumRecordingTime, to allow a user to set a desired maximum length of any transmission file.

EricTendian:
This implements enhancement #197 - if the length of call audio exceeds a user-specified duration, it will end the call and start a new one where it left off. If no duration is specified it will default to 0 (the current behavior). To clear up the confusion from #197, the duration is measured by the length of active transmission/audio, not real-time.

I didn't actually implement the change (see the co-author), but I've been using it for a couple months now on https://openmhz.com/system/chi_cpd to split up long calls.

Fixes #197

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,6 +139,7 @@ Here are the different arguments:
  - **defaultMode** - Default mode to use when a talkgroups is not listed in the **talkgroupsFile** the options are *digital* or *analog*.
  - **captureDir** - the complete path to the directory where recordings should be saved.
  - **callTimeout** - a Call will stop recording and save if it has not received anything on the control channel, after this many seconds. The default is 3.
+ - **maxDuration** - If a call is being recorded and the duration exceeds this value in seconds, the call will stop recording and a new call will pick up where the previous on left off. The default is *0* or "no timeout".
  - **logFile** - save the console output to a file. The options are *true* or *false*, without quotes. The default is *false*.
  - **frequencyFormat** - the display format for frequencies to display in the console and log file. The options are *exp*, *mhz* & *hz*. The default is *exp*.
  - **controlWarnRate** - Log the control channel decode rate when it falls bellow this threshold. The default is *10*. The value of *-1* will always log the decode rate.
--- a/trunk-recorder/config.cc
+++ b/trunk-recorder/config.cc
@@ -140,6 +140,8 @@ Config load_config(std::string config_file, std::vector<Source *> &sources, std:
     BOOST_LOG_TRIVIAL(info) << "Log to File: " << config.log_file;
     config.control_message_warn_rate = pt.get<int>("controlWarnRate", 10);
     BOOST_LOG_TRIVIAL(info) << "Control channel rate warning: " << config.control_message_warn_rate;
+    config.max_duration = pt.get<int>("maxDuration", 0);
+    BOOST_LOG_TRIVIAL(info) << "Maximum Call Duration (seconds): " << config.max_duration;
 
 
     BOOST_FOREACH(boost::property_tree::ptree::value_type  & node,
--- a/trunk-recorder/config.h
+++ b/trunk-recorder/config.h
@@ -26,6 +26,7 @@ struct Config {
         int call_timeout;
         bool log_file;
         int control_message_warn_rate;
+        int max_duration;
         int control_retune_limit;
         bool broadcast_signals;
 };
--- a/trunk-recorder/main.cc
+++ b/trunk-recorder/main.cc
@@ -543,6 +543,8 @@ void load_config(string config_file)
     BOOST_LOG_TRIVIAL(info) << "Control channel warning rate: " << config.control_message_warn_rate;
     config.control_retune_limit = pt.get<int>("controlRetuneLimit", 0);
     BOOST_LOG_TRIVIAL(info) << "Control channel retune limit: " << config.control_retune_limit;
+    config.max_duration = pt.get<int>("maxDuration", 0);
+    BOOST_LOG_TRIVIAL(info) << "Maximum Call Duration (seconds): " << config.max_duration;
 
     BOOST_LOG_TRIVIAL(info) << "Frequency format: " << frequencyFormat;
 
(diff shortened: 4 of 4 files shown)


bctrainers:
I haven't merged and compiled this pull locally, but does the max duration perform a hard cut at X seconds configured (thus cutting off active audio)? Or is it a soft cut, where the next keyup that happens past the max duration starts a new recording?

EricTendian:
@bctrainers it's a hard cut, though when you listen to it back to back, you don't notice it (apart from maybe a slight delay if the second file needs to be downloaded). The OpenMHz system I linked above has a couple cases of it (just look for the recordings of 60 or 61 seconds).

bctrainers:
Would it be possible to have the code check the call status rather than doing hard-timed cuts? Only restart the recorder if the Radio ID changes at maximum duration or exceeding the timer set. So when the next RID keys up, it's off to a new recorder.

On the other hand, this sort of would end up in a feature request realm, as this code appears to also be for all captures (analog and digital) rather digital/P25. Having a "HardMaxDuration" with the current code and a "DigitalMaxDuration" for trunking systems might work in this instance. Where DigitalMaxDuration would work coinciding with HardMaxDuration (open/hot mic instances).

EricTendian:
@bctrainers yeah, that might be more of a separate request, due to the complexities you mentioned. #326 sounds like it aligns more closely with what you want, or at least the logic needed for implementing that could probably be applied.

robotastic:
@EricTendian It looks like this change would only work for conventional system. From #197 it sounds like it could be a problem with trunked systems too. I haven't seen it on the systems I monitor though. I am happy to add it in just for conventional, could you update the Readme though to say it is only for conventional?

@bctrainers I started looking into this to try and address #326 but it got really complicated quickly. Since you have multiple Grant messages repeated, the logic is messy. The other problem is unless you push some dead air time though the recorder, I think there maybe some audio left in the buffers. It maybe possible to listen for a Termination message on P25 audio channel and then pipe that back to the main control loop... but that seems like it may introduce more problems than it solves. Shortening the Idle time is probably the best way to make sure a long conversations are not started.

EricTendian:
@robotastic I have this working on a trunked system as well, check out the 61 second recording at 6:28pm on https://openmhz.com/system/sc21102?filter-type=group&filter-code=5a499409129ce10001425ba0. It restarted the call after 60 seconds. There's a few other examples of 60 second recordings if you scroll back far enough on that link.

dkettmann:
I found my version of the code from the work done on #197  .  I am 100% noobing it up, but I think to make this work for digital we ended up doing the following also.  Otherwise this looks spot on with everything else.  Thank you so much for making this issue, I'm clueless with all this. 
dan@bmetalcity18:~/trunk-recorder$ diff main.cc main.cc.orig
767c767
<       if (call->since_last_update() > config.call_timeout || ( call->get_current_length() > config.max_duration && config.max_duration > 0 )) {
---
>       if (call->since_last_update() > config.call_timeout) {

(Took me *hours* to learn git enough to do the following to pull down a commit.)
git remote add origin https://github.com/robotastic/trunk-recorder.git
git fetch origin
git fetch origin 7fe3477410dc0ab4146fbefb6e0afe3bd85f5209
git reset --hard FETCH_HEAD

dkettmann:
I found my version of the code from the work done on #197. I am 100% noobing it up, but I think to make this work for digital we ended up doing the following also. Otherwise this looks spot on with everything else. Thank you so much for making this issue, I'm clueless with all this.

```
dan@bmetalcity18:~/trunk-recorder$ diff main.cc main.cc.orig
767c767
<       if (call->since_last_update() > config.call_timeout || ( call->get_current_length() > config.max_duration && config.max_duration > 0 )) {
---
>       if (call->since_last_update() > config.call_timeout) {
```

(Took me hours to learn git enough to do the following to pull down a commit.)
git remote add origin https://github.com/robotastic/trunk-recorder.git
git fetch origin
git fetch origin 7fe3477
git reset --hard FETCH_HEAD

bctrainers:
> I found my version of the code from the work done on #197. I am 100% noobing it up, but I think to make this work for digital we ended up doing the following also. Otherwise this looks spot on with everything else. Thank you so much for making this issue, I'm clueless with most of this.
> 
> ```
> dan@bmetalcity18:~/trunk-recorder$ diff main.cc main.cc.orig
> 767c767
> <       if (call->since_last_update() > config.call_timeout || ( call->get_current_length() > config.max_duration && config.max_duration > 0 )) {
> ---
> >       if (call->since_last_update() > config.call_timeout) {
> ```
> 
> (Took me hours to learn git enough to do the following to pull down a commit.)
> 
> ```
> git remote add origin https://github.com/robotastic/trunk-recorder.git
> git fetch origin
> git fetch origin 7fe3477410dc0ab4146fbefb6e0afe3bd85f5209
> git reset --hard FETCH_HEAD
> ```

I will poke about with this diff on my local copy and see how things end up.

PR #358, opened 2020-05-19
Clear analog source list on call end, adding support for P25 conventional source ID logging

### LINKED ISSUE #142: Motorola Data Communications Decoding (Stat-Alert, MDC-600, MDC-1200)
# Background
MDC uses Audio Frequency Shift Keying (AFSK) at a 600 or 1200 baud rate. AFSK digital data is represented by changes in the frequency (pitch) of an audio tone. With this scheme, the "1" is called the mark frequency (`1200 Hz`) and the "0" is called the space frequency (`1800 Hz`). (AFSK is also used in [EAS](https://en.wikipedia.org/wiki/Emergency_Alert_System) to notify stations of the type of emergency, locations affected, and the time of issue without actually hearing the text of the alert. Caller ID on phone lines also uses the AFSK method.)

# Uses
Data bursts can be send pre or post audio transmission.
**Pre-Tone** data bursts sent when the user presses the PTT button and there is a short delay before the user can speak. Useful as it allows the receiver to see the PTT-ID of the currently speaking user. 
**Post-Tone** Data bursts sent out after the user depresses the PTT button. Uses slightly less air time than the pre-tone variant.
* PTT-ID (sometimes called ANI)
  * Are Hexadecimal [0-9 & A-F] four-digit numbers. (Ex `0000`, `FFFF`, `2900`, `A291`, `291C`, etc).
* Emergency Button
  * Sends the PTT-ID 3 times with no audio.
* Status Buttons
  * Status 1 - Status 8 are available. Protocol implementation is unknown to me.
* Selective Inhibit
  * "Stuns" a radio that might be stolen, or the user keeps an open carrier and makes it unable to transmit. Protocol implementation is unknown to me.
* Radio Check
  * If the radio is in range and turned on, after sending the radio check, the destination radio will respond with an ack. Protocol implementation is unknown to me.
* Selective Calling
  * If the radio is in range and turned on, and receives a selective calling message, the radio could display a "CALL" message and / or make an audible BEEP. 

# Implementation
With the P25 decoder, unit ID information is passed out into the json file. Ideally the same information would be present on conventional systems as well. This would make feature parity between the currently supported implementations.

### LINKED ISSUE #307: Unit IDs for conventionalP25?
My FD just switched to a conventional P25 system, I started running OP25 for live audio and I noticed that the Radio IDs do come up just like with trunked P25 systems. Is there an option currently to output these to the json file in trunk-recorder like with smartzone/p25 systems?

EricTendian:
This PR does pretty much what the title says:

1. The analog source logging wasn't getting cleared on call end, so they kept piling up. This fixes that bug.
2. While there was source logging for P25 trunking systems, there wasn't for P25 conventional systems - this adds that feature.

The changes are a subset of those in #322.

I've tested this on an analog system as well as a P25 conventional one and verified on both the source ID logging works as expected, starting fresh for each new call.

I did have to make https://github.com/robotastic/trunk-recorder/commit/4365f1b7459eb612f241437b513a970839960889 to fix a compile error introduced by https://github.com/robotastic/trunk-recorder/commit/a7a645ca7f9f096563ac8da616b58d50609571c6.

Fixes #307
Fixes #142

### DIFF
--- a/lib/gr_blocks/nonstop_wavfile_sink.h
+++ b/lib/gr_blocks/nonstop_wavfile_sink.h
@@ -72,6 +72,9 @@ public:
 	 */
 	virtual void set_bits_per_sample(int bits_per_sample) = 0;
 	virtual double length_in_seconds() = 0;
+
+	virtual void set_call(Call* call) {};
+	virtual void end_call() {};
 };
 
 } /* namespace blocks */
--- a/lib/gr_blocks/nonstop_wavfile_sink.h
+++ b/lib/gr_blocks/nonstop_wavfile_sink.h
@@ -74,7 +74,6 @@ public:
 	virtual double length_in_seconds() = 0;
 
 	virtual void set_call(Call* call) {};
-	virtual void end_call() {};
 };
 
 } /* namespace blocks */
--- a/lib/gr_blocks/nonstop_wavfile_sink_impl.cc
+++ b/lib/gr_blocks/nonstop_wavfile_sink_impl.cc
@@ -213,6 +213,25 @@ bool nonstop_wavfile_sink_impl::stop()
   return true;
 }
 
+void nonstop_wavfile_sink_impl::log_p25_metadata(long unitId, const char* system_type, bool emergency)
+{
+  if (d_current_call == NULL) {
+    BOOST_LOG_TRIVIAL(debug) << "Unable to log: " << system_type << " : " << unitId << ", no current call.";
+  }
+  else {
+    BOOST_LOG_TRIVIAL(debug) << "Logging " << system_type << " : " << unitId << " to current call.";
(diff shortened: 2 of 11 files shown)


robotastic:
It doesn't look like end_call() gets called anywhere. Is this function needed? Could _current_call be set to null when a new file is opened and/or at close()?

EricTendian:
@robotastic oh, good point, I didn't realize that - I just updated it to set the current call to null when the file is closed.

robotastic:
Looks good - I added a bonus NULL at the beginning just to be sure. Did you have a chance to make sure it also works with P25Conventional when you are using delayed open? It looks like that shouldn't be a prob because the wav-sink do_work function is still being called by the delayedopen_wavsink and the class is built on wavsink.

EricTendian:
@robotastic thanks! I didn't test this with delayed open, but you're right that based on how the logic is implemented it seems like it wouldn't be an issue.

PR #374, opened 2020-06-19
Updated audioplayer.php

jquagga:
So I figured I would pass this along for consideration.  I don't think the existing audioplayer.php works with the semi-standard configuration of trunk-recorder these days.  I did some modifications to get it to handle multiple systems and modify the $files array to not use mtime to get the script showing all of my entries.  I bumped bootstrap to 4.5 and went with the media trick to have a "dark mode".  That part may not be desirable as it changes the overall appearance of the page to a more standard black text on white (or white text on black).  But I liked it.  

This should work from the trunk-recorder directory with the standard ```php -S 127.0.0.1:8080``` option and I also have a docker-compose I can share which lets it use nginx / php-fpm to interact with the trunk-recorder-docker setup and serve out those files.  

I know it's no longer "the way" but I've wanted this working for a long time as I think it's a handy, lightweight way of checking on what my system is "hearing" without logging into one of the services.  

screenshot-fs8

-----

* Updating the player to support multiple systems / the new media path.
* Index files by time and frequency, rather than on MTIME (which can be the same for several files causing them to be invisible)
* Update bootstrap to 4.5 and configure flatly / darkly for a bright and night mode to match the browser
* Link to the m4a files instead of the wav files
* php linter changes

### DIFF
--- a/audioplayer.php
+++ b/audioplayer.php
@@ -68,6 +68,7 @@ try {
     <head>
         <meta charset="utf-8">
         <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
+        <meta name="format-detection" content="telephone=no">
         <title>Trunk Player</title>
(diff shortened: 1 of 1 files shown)


robotastic:
Thanks @bctrainers & @jquagga - looks like an improvement!

Issue #365, opened 2020-06-05
Docker status


robbiet480:
Hi @robotastic,

Thought I should create a issue to continue work from https://github.com/robotastic/docker-gnuradio/issues/2. Now that I've figured out the path to get multiple architectures building properly, i'm working on improving the entire `trunk-recorder` Docker ecosystem. This includes updating the two repos `Dockerfile` to follow best practices, slimming them down as much as possible and improving the Docker Compose story. Unrelated to `trunk-recorder` specifically, I'm also working on containerizing @chuot's `rdio-scanner` so that users can get a whole stack running with a single Docker Compose file and much faster too, especially for those on Raspberry Pi. 

I've created this issue because I was hoping to get the current lay of the land from you and @EricTendian. I see a lot of Docker issues in multiple repos but am unsure where my time is best spent so as to not duplicate work. So, a few questions to start:

1. Based on #349 it appears we need a new Dockerfile to support GNURadio 3.8. Is this correct? I see https://github.com/robotastic/trunk-recorder-docker-gr3.8, is this actually working? Should we start using GNURadio 3.8 as the base for Docker?
2. What's the issue with Locales? I see you are setting everything to `en_US.UTF-8`. Why? Can we stop doing that and just use `C` or `C.UTF-8`?
3. Is there any way to build a static binary so that we can just have the GNURadio and other deps from `docker-gnuradio` and none of the extra cruft of this repo? I like keeping Docker images nice and small so its quick for users to get going.
4. Are there any magic cmake or make flags I can set to optimize cross compiling? As mentioned in https://github.com/robotastic/docker-gnuradio/issues/2 compile times are quite long right now. I'd like to speed those up if possible.

Here's the current Dockerfile I'm using which works nicely and hews to expected Docker standards:

```Dockerfile
FROM robotastic/docker-gnuradio:latest

WORKDIR /src

COPY . .

RUN cmake . && make -j`nproc` && cp recorder /recorder

CMD ["/recorder"]
```

And the new and improved `docker-compose.yaml`, featuring `rdio-scanner` support too:

```yaml
version: '3'
volumes:
  rdio_database:
services:
  recorder:
    image: robotastic/trunk-recorder:latest
    container_name: trunk-recorder
    restart: always
    devices:
      - /dev/bus/usb/001/003:/dev/bus/usb/001/003
    volumes:
      - ./media:/app/media
      - ./config:/app/config
      - ./ebrcs.csv:/app/ebrcs.csv
[...]

PR #375, opened 2020-06-21
Docker: Set the config file path

robbiet480:
Fixes issue realized by @jquagga in #365. Once merged, I’ll update the `docker-compose.yml` example in the wiki too.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,4 +8,4 @@ RUN cmake . && make -j`nproc` && cp recorder /recorder
 
 USER nobody
 
-CMD ["/recorder"]
+CMD ["/recorder", "--config=/app/config.json"]


jquagga:
Greetings,

Sorry, I'm perpetually behind trying these out.  Today I tried to use the docker-compose from the wiki and the latest "nightly" which has this fix applied.  I'm still getting the config path being slightly off:

```
Starting trunk-recorder ... done
Attaching to trunk-recorder
trunk-recorder | [2020-06-25 11:41:35.044531] (info)   Using Config file: /app/config.json
trunk-recorder | [2020-06-25 11:41:35.044775] (error)   Failed parsing Config: /app/config.json: cannot open file
```

Comparing the [Dockerfile in the trunk-recorder repository](https://github.com/robotastic/trunk-recorder/blob/master/Dockerfile) with the [older Dockerfile in the trunk-recorder-docker repository](https://github.com/robotastic/trunk-recorder-docker/blob/master/Dockerfile) (which is where the "legacy" docker-compose file comes from) the path's on the --config are slightly different.  

I believe the wiki example docker-compose.yml is based on having ```CMD ["./recorder","--config=/app/config/config.json"]``` like the legacy Dockerfile but the current Dockerfile has ```CMD ["/recorder", "--config=/app/config.json"]```  It's missing the "config" directory in the middle.  I think the Dockerfile or the docker-compose.yml example on the wiki could change; it doesn't matter either way.  

So I worked around that for now in the docker-compose.yml to see how far I could test this.  The next challenge appeared to be the .csv talkgroup files.  trunk-recorder couldn't find them in ```/app```.  I thought they should be in / since recorder is now running as ```/recorder``` but that didn't work either.  It seems it does work if they are set to ```/src/ChanList.csv```.  That would let trunk-recorder find the .csv file and almost work. 

The last challenge was that running as nobody, it doesn't seem to have the ability to fully function when running as User nobody.  The image comes up, but hangs before locking onto a control channel.  If you run the same image (nightly) as user root:root, it works fine. I tested this another way - presently the difference between :nightly and :edge appears to be the User nobody in the Dockerfile.  :edge works fine, :nightly doesn't until you add the user root:root to the docker-compose.yml.  I'm not really sure why.  The rtlsdr's are both a+rw so I would have though nobody could access them either way.  

Right now, this is the docker-compose.yml I have on my test system and it appears to be working with :edge or nightly (if the user root:root is there for nightly).  

```
version: '3'
services:
  recorder:
    image: robotastic/trunk-recorder:edge
    container_name: trunk-recorder
    restart: unless-stopped
    devices:
      - /dev/bus/usb:/dev/bus/usb
    volumes:
      - ./media:/app/media
      - ./config/config.json:/app/config.json
      - ./ChanList.csv:/src/ChanList.csv
      - ./PWCChanList.csv:/src/PWCChanList.csv
    user: root:root
```

PR #377, opened 2020-06-24
Correct TGID lookup in audioplayer.php

jquagga:
I know that audioplayer.php probably isn't the most used piece of code in the repository and I was going to leave it alone after the last PR, but I found a bug.  I think I squashed it locally so figured I'd pass this along.  

While the file lookups work fine with the last update, the TGID updates weren't separated by system.  So if multiple systems are in place, and they had overlapping decimal talkgroups whichever system was loaded last got the correct alphatag.  On my two system recorder this lead to many talkgroups being correct and some being completely mislabelled.  

* The major part of this update separates $TGS by $SHORTNAME.  $TGS becomes a multi-dimensional array of [$SHORTNAME][TGID] and the lookup is changed to match.  Basically we have an alphatag lookup of talkgroup for each system listed in the config.

There are a few other updates to the file as a result of html5 validation which I can remove if you'd like:

* lang is set in the <html>
* The <style> element moves into the <head> from <body>

And finally there are a couple of cosmetic updates which are completely optional:

* There is now a "jump to bottom" button which scrolls the page to the bottom (the most recent entries).
* I changed the bootstrap colors to primary which seems to work better when flipping themes for day/night mode.

screenshot-fs8

### DIFF
--- a/audioplayer.php
+++ b/audioplayer.php
@@ -28,9 +28,11 @@ $TGFile = function (string $tgFilePath): array {
     return $return;
 };
 
-$TGS = [];
+
 foreach ($CONFIG->systems as $system) {
-    $TGS += $TGFile($system->talkgroupsFile);
+    $SHORTNAME = "{$system->shortName}";
+    $TGS[$SHORTNAME] = [];
+    $TGS[$SHORTNAME] += $TGFile($system->talkgroupsFile);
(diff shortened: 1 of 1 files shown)


robotastic:
Thanks for keeping this up to date!

PR #378, opened 2020-07-09
Update auto-restart.sh

e850205:


robotastic:
Thanks! Prob good to have that in.

PR #381, opened 2020-07-24
Make defaultMode only affect smartnet systems

leee:
## Summary

P25 trunks don't use analog talkpaths. Therefore, `defaultMode` should only
steer `smartnet` systems. Since there are 4 cases where only 1 is affected:

```
- TII, default ana -> record analog
- TII, default dig -> record digital
- P25, default ana -> record digital (ignore defaultMode)
- P25, default dig -> record digital (ignore defaultMode)
```

this pull request adds `get_system_type() == "smartnet"` with the existing
`default_mode == "analog"` test.

I can't speak authoritatively for P25 trunks of all manufacturers and haven't
read specs in ages, but can say that ASTRO25 systems (and subscribers I've
experience with) don't support analog. This also includes failsoft operation -
only TII lets you choose failsoft signal type and secure operation, and that's
done on a subscriber basis. In normal operation on TII, both subscriber
programming and system talkgroup configuration must be in lockstep.

`defaultMode` now defaults to `analog`. This is potentially a breaking change
for existing configurations that listen to unknown TII trunks with only digital
use, and have not included `defaultMode` in their configuration. Users may need
to be informed on update.

I haven't thoroughly familiarized myself with the codebase, but my quick read
over `main.cc` is that `find/get_talkgroup` is used in trunking calls only, so
this change shouldn't affect any `conventional`/`conventionalP25` systems.

## Improvements

In a better world, `defaultMode` should be a system key and not a global knob.
`trunk-recorder` should respect differing `defaultMode` for separate TII trunks.
A default `defaultMode` of `digital` would work fine for trunks that provision
all their talkgroups as digital, but such a system is an exception and not the
norm. `0A08` is the only trunk I can think off the top of my head that was a TII
trunk that was mostly (if not exclusively) ASTRO.

In an ideal world, handling of channel grant status bits should be implemented
and `defaultMode` be retired. Unknown talkgroups can potentially be used in
either mode. Even with known fleetmapped talkgroups already provisioned as one
or the other, and TII trunks being legacy with changes likely not happening in
this day and age, it's still plausible that trunks on life support can add
additional digital talkpaths and change the mode of existing talkgroups.

These are improvements for consideration that should be created as a new issue,
but I'm not familiar with the social norms of this project to know if that's
acceptable or not.

## Testing

This PR has been tested working with the following:

```
$ uname -a
Linux ubuntu 5.3.0-53-generic #47~18.04.1-Ubuntu SMP Thu May 7 13:10:50 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux

$ apt list ... # not included since this was a small change
```

```
$ cat test-config.json
{
  "sources": [
    {
      "center": ,
      "rate": ,
      "squelch": -55,
      "error": 0,
      "gain": 15,
      "digitalRecorders": 32,
      "digitalLevels": 8,
      "modulation": "qpsk",
      "analogRecorders": 16,
      "analogLevels": 8,
      "debugRecorders": 0,
      "driver": "osmosdr",
      "device": "rtl=n,buflen=65536"
    }
  ],
  "systems": [
    "irrelevant conventional systems elided",
    {
      "shortName": "SYS.SITE",
      "type": "smartnet",
      "control_channels": [],
      "bandplan": "",
      "bandplanBase": ,
      "bandplanHigh": ,
      "bandplanSpacing": ,
      "bandplanOffset": ,
      "talkgroupsFile": "empty.csv"
    },
    {
      "shortName": "WACN-SYS.RFSS-SITE",
      "type": "p25",
      "control_channels": [],
      "talkgroupsFile": "empty.csv",
    }
  ],
  "defaultMode": "analog"
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,7 +136,7 @@ Here are the different arguments:
    - **decodeFSync** - *(Optional, For conventional systems)* enable the Fleet Sync signaling decoder. The options are *true* or *false*, without quotes. The default is *false*.
    - **decodeStar** - *(Optional, For conventional systems)* enable the Star signaling decoder. The options are *true* or *false*, without quotes. The default is *false*.
    - **decodeTPS** - *(Optional, For conventional systems)* enable the Motorola Tactical Public Safety (aka FDNY Fireground) signaling decoder. The options are *true* or *false*, without quotes. The default is *false*.
- - **defaultMode** - Default mode to use when a talkgroups is not listed in the **talkgroupsFile**. The options are *digital* or *analog*. The default is *analog*. This argument is global and not system-specific, and only affects `smartnet` systems as `p25` systems don't have analog talkpaths.
+ - **defaultMode** - Default mode to use when a talkgroups is not listed in the **talkgroupsFile**. The options are *digital* or *analog*. The default is *digital*. This argument is global and not system-specific, and only affects Type II `smartnet` trunking systems which can have both analog and digital talkpaths whereas `p25` trunking systems don't have analog talkpaths.
  - **captureDir** - the complete path to the directory where recordings should be saved.
  - **callTimeout** - a Call will stop recording and save if it has not received anything on the control channel, after this many seconds. The default is 3.
  - **logFile** - save the console output to a file. The options are *true* or *false*, without quotes. The default is *false*.
--- a/README.md
+++ b/README.md
@@ -136,7 +136,7 @@ Here are the different arguments:
    - **decodeFSync** - *(Optional, For conventional systems)* enable the Fleet Sync signaling decoder. The options are *true* or *false*, without quotes. The default is *false*.
    - **decodeStar** - *(Optional, For conventional systems)* enable the Star signaling decoder. The options are *true* or *false*, without quotes. The default is *false*.
    - **decodeTPS** - *(Optional, For conventional systems)* enable the Motorola Tactical Public Safety (aka FDNY Fireground) signaling decoder. The options are *true* or *false*, without quotes. The default is *false*.
- - **defaultMode** - Default mode to use when a talkgroups is not listed in the **talkgroupsFile** the options are *digital* or *analog*.
+ - **defaultMode** - Default mode to use when a talkgroups is not listed in the **talkgroupsFile**. The options are *digital* or *analog*. The default is *analog*. This argument is global and not system-specific, and only affects `smartnet` systems as `p25` systems don't have analog talkpaths.
  - **captureDir** - the complete path to the directory where recordings should be saved.
  - **callTimeout** - a Call will stop recording and save if it has not received anything on the control channel, after this many seconds. The default is 3.
  - **logFile** - save the console output to a file. The options are *true* or *false*, without quotes. The default is *false*.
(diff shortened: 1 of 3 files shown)


robotastic:
Let me go double check. It has been a while but if I remember right P25 allows for analog voice channels in Phase 1. Not sure if that is used anywhere. Making this a per-system setting might be the cleanest fix.

The `defaultMode` should stay defaulted to digital, since most system the TR is used with are digital systems. Adding a talkgroup file is an optional feature.

In many cases it is easier to run a separate instance of trunk recorder for each system you are trying to record... unless they are overlapping and a single SDR can cover both.

leee:
> It has been a while but if I remember right P25 allows for analog voice 
> channels in Phase 1.

Perhaps you're confusing this for subscriber or FNE capability. Subscribers can
(and should) have TIA-603 (FM/PM analog air interface) capability, and a
manufacturer can give you FNE designed for trunking that can do analog (such as
conventional analog or trunking analog, generally to prepare you for a system
upgrade). This is a compat/interoperability thing, and TIA-603 is not used in
P25 trunking.

The other thing I think you could be confusing this with is when documentation
or specs refer to "Project 25 Phase 0". Phase 0 was meant to refer to non-P25
standards for analog systems (conventional or otherwise). This by definition is
not P25, and Project 25 Phase 1 and 2 both require a use of a digital CAI.

> Making this a per-system setting might be the cleanest fix.

I agree, and this PR was meant to just be a quick sensibility fix. It doesn't
make sense to record a P25 trunking system that will never be on the air in
analog mode.

I hope to make a lot of things per-system, including modulation.

> The defaultMode should stay defaulted to digital, since most system the TR is
> used with are digital systems. Adding a talkgroup file is an optional feature.

If this change is implemented, then P25 trunks will remain recording in digital.
Type II trunks will of course record in analog unless this default is changed to
digital.

> In many cases it is easier to run a separate instance of trunk recorder for
> each system you are trying to record... unless they are overlapping and a
> single SDR can cover both.

A common case might be a legacy Type II trunk that is being replaced with P25
trunking and is reusing the same frequency pairs. Since you already have
hardware in place to record this set of output channels, it seems sense that you
use the same hardware to begin recording the new system as it comes on the air.

Another case might be systems that cover the same geographic area but have
different owners, or different coverage plans. A theoretical example can be a
location having a state-run trunking site, a county-run trunking site, and a
city-owned trunking site. A latter example could be a trunking system using the
same band for all sites, but one site is meant for mobile wide-area coverage
and another meant for portable coverage.

bctrainers:
If I recall correctly, most P25 radios have a last-resort fallback to conventional/analog in the event all primary and alternative control channels are exhausted and are unable to roam to another nearby system (if that has been configured). However, this fallback is incredibly rare (I've only seen this happen once here in my region, and that was when the radio techs applied a bad configuration to the metros P25 system) :-)

From checking the code, all looks well. Interesting comments contained therewithin. 👍

leee:
> If I recall correctly, most P25 radios have a last-resort fallback to conventional/analog in the event all primary and alternative control channels are exhausted and are unable to roam to another nearby system (if that has been configured). However, this fallback is incredibly rare (I've only seen this happen once here in my region, and that was when the radio techs applied a bad configuration to the metros P25 system) :-)

Indeed, "degradation" or "fallback" mode is a mandatory requirement in Project 25. A system can gracefully degrade to single site mode (in Motorola/ASTRO25 parlance: "site trunking"), or if the site really wants to kick techs in the shin, degrade down to the next level of conventional fallback (A25: "failsoft").

To be pedantic, IIRC, site trunking is a mandatory option whereas conventional fallback is a "standard option", chosen to be implemented or not by the customer (but since it's "free", they should). All talkpaths remain digital.

robotastic:
I just skimmed the specs and it doesn't look like P25 control channel messages support analog channels. Having `default_mode` only apply for SmartNet makes sense. 

Most SmartNet systems I have seen are digital, so I would like the default configuration for `default_mode` to remain digital.

leee:
No problem, we can keep it digital for your experience.

PR #388, opened 2020-08-03
Type II encryption and emergency group call handling

leee:
Not much to report here. This PR adds parsing of the status nibble for encrypted and emergency signaling.

Good thing here means if the talkgroup grant/continue is coming up as encrypted, `main.cc` doesn't bother recording it.

One thing to particularly note is we're still using `full_address` for the talkgroup. I'm not willing to switch that to just `talkgroup` right now, because that will be a breaking change and needs to have some discussion in an issue for coordination before making and merging that change.

### DIFF
--- a/trunk-recorder/systems/smartnet_parser.cc
+++ b/trunk-recorder/systems/smartnet_parser.cc
@@ -238,8 +238,15 @@ std::vector<TrunkMessage> SmartnetParser::parse_message(std::string s,
       message.message_type = UPDATE;
       message.freq = getfreq(stack[3].cmd, system);
       message.talkgroup = stack[3].full_address;
-      // message.encrypted    = false;
-      // message.emergency    = false;
+      if (stack[3].status >= 8) {
+        message.encrypted = true;
+        if ((stack[3].status == 10) || (stack[3].status == 12) || (stack[3].status == 13)) {
+          message.emergency = true;
+        }
+      }
+      if ((stack[3].status == 2) || (stack[3].status == 4) || (stack[3].status == 5)) {
+        message.emergency = true;
+      }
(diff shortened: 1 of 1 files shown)


trunk-recorder/systems/smartnet_parser.cc
```diff
@@ -338,8 +345,15 @@ std::vector<TrunkMessage> SmartnetParser::parse_message(std::string s,
         message.message_type = GRANT;
         message.freq = getfreq(stack[2].cmd, system);
         message.talkgroup = stack[2].full_address;
-        // message.encrypted    = false;
-        // message.emergency    = false;
+        if (stack[3].status >= 8) {
```
bctrainers, line 348: Is it wise to assume a greater-than or equal to clause in this instance? 

While Moto SmartNet Type 2 likely won't have any additional status-bits added any time soon, this could be a remote-chance problem in the future if more bit types were to be added or amended.

bctrainers: Same applies to: https://github.com/robotastic/trunk-recorder/pull/388/files#diff-4a7f416921ca59465ac64f045a1d86f5R241  :)

leee: The status nibble is just that, 1 hex digit, part of a grant or continue OSW. Most of the values (all 16 but 2) are known/defined, with the upper half being encrypted and the lower 8 being not encrypted. A `>= 8` test covers this formulation.

robotastic:
This looks good to me. Could you create an issue regarding `full_address` I still haven't seen good documentation on how addressing works in SmartNet. Could you describe what needs to be changed?

PR #395, opened 2020-08-17
Basic build improvements for Ubuntu 20.04 and minor cosmetics

aSmig:
This should move things in the direction of just running ./install.sh on a fresh minimal install of Ubuntu 20.04 LTS.

The logging changes are super insignificant, but I like them. Feel free to drop these changes.


robotastic:
Thanks! These look good to me. Merging in

PR #400, opened 2020-09-19
Ignore status bits when computing Type II talkgroup address

### LINKED ISSUE #389: Talkgroup Numbering
Currently, `trunk-recorder` (as well as OpenMHz) tries to operate on the common ground of talkgroups being in decimal and ignores hexadecimal talkgroup values. This isn't a poor choice, since it is my understanding that the various scanner websites (and by extension, the scanner hardware they support) operate in decimal as well.

This works, so long as everything stays on the same page. Pages include:
- `trunk-recorder` writes to files prefixed with talkgroup in decimal
- talkgroups are handled on OpenMHz in decimal
- talkgroup `csv`s for OpenMHz are handled in decimal only as well (and intentionally requires that you remove the hexadecimal talkgroup column, occasionally a pain point seen in issues like #266)

In theory, local talkgroup `csv`s for recorders should be handled in decimal as well, but you occasionally see things not fall in lockstep like in #306 which affects P25 systems:

> After digging around, folks on gitter chat concluded that the HEX value in the CSV has to be filled out properly for the priority to function.

---

Type II trunks make this hex vs dec juggling worse. Hexadecimal talkgroups are joined with a status nibble in the control channel stream. So instead of "TG `$ABC`", you see "Data field `$ABCD`", where `$D` is that status nibble whose value tells us if a talkgroup is encrypted, if there's an emergency call on it, if it's a patch or msel, if it's an ATG, etc. We shouldn't be logging calls like this, which should be handled by just talkgroup and not talkgroup with status.

This is compounded by the scanner website data. Such data is always saved with decimal talkgroups always having `0` status. Calls with a non-0 status nibble do not get matched to these known talkgroups.

Now that we are aware of encrypted calls and emergency calls with #388, we need to start thinking about how to fix this issue so non-0 status talkgroups don't get mishandled like they are right now.

---

As we talk about this though, I find it important to bring up the fact that trunks frequently get upgraded piece by piece, and talkgroups remain the same as legacy infrastructure gets homed on P25 infrastructure. That means a talkgroup file for a Type II trunk is also good for the P25 trunk replacing it.

However, we currently would not be able to use the same `csv` for both a Type II trunk and a P25 trunk. Since we try to operate on decimal, and Type II trunks include the status nibble, decimal talkgroups do not line up and are bit-shifted by 4 thanks to that status nibble! We frequently see this "Type II scanner land decimal talkgroup" to "actual decimal talkgroup, also used by P25" as a multiple of 16.

What does line up though are hex talkgroups. For some reason, the scanner websites have deemed it useful for Type II listings to have decimal and hexadecimal talkgroup columns, but the talkgroup columns don't retain a status nibble, whereas decimal columns do.

---

Since we'd have a breaking change trying to fix Type II talkgroup numbering, we may as well take advantage of the timing and make everything operate on hexadecimal talkgroup numbering. In order to do this though, several things will have to happen:

1. ensure popular upload targets like OpenMHz are prepared for hexadecimal talkgroup formatted audio and json file names - this could be by including a `v2` or `all hex` somewhere on upload, or similar notification.
2. think about users in this change. Users can come in several flavors for this change, and I'll discuss that below.
3. make all parts of `trunk-recorder` operate on hexadecimal talkgroups and upload targets be updated with a v2 flag.
4. include this change only in a major release where users are hopefully well informed that they need to double check their config and talkgroup files.

Regarding users: they can come in the following flavors:
1. local talkgroup files with hex columns populated correctly. All they have to do is update their targets, especially ones like OpenMHz which required stripping that column before.
2. local talkgroup files with hex columns not populated. We can alert but continue to operate as if we had no local TG file. The hope is that any messaging and notices upon updates is what alerts these users to update their files.
3. local talkgroup files with hex columns filled with garbage data. The easy way out would be to respond in kind: garbage in means garbage out.

It's possible that this change could involve a configuration setting for operating in hexadecimal or decimal (but which kind, real or scanner-land for Type II systems?) but the more I think about it, the more I feel that it's easiest to rip out the pain point in one go than apply bandages that will need to be handled eventually, which is just kicking the problem down the road.

ginomiglio:
I just took @robotastic's `address` branch and rebased it onto latest master. Partial fix for #389.

### DIFF
--- a/trunk-recorder/systems/smartnet_parser.cc
+++ b/trunk-recorder/systems/smartnet_parser.cc
@@ -237,7 +237,7 @@ std::vector<TrunkMessage> SmartnetParser::parse_message(std::string s,
       //     << std::hex << stack[4].cmd << " " << std::hex << stack[4].grp << " " << std::hex << stack[4].full_address << " ]";
       message.message_type = UPDATE;
       message.freq = getfreq(stack[3].cmd, system);
-      message.talkgroup = stack[3].full_address;
+      message.talkgroup = stack[3].address;
       if (stack[3].status >= 8) {
         message.encrypted = true;
         if ((stack[3].status == 10) || (stack[3].status == 12) || (stack[3].status == 13)) {
@@ -344,7 +344,7 @@ std::vector<TrunkMessage> SmartnetParser::parse_message(std::string s,
         //     << std::hex << stack[4].cmd << " " << std::hex << stack[4].grp << " " << std::hex << stack[4].full_address << " ]";
         message.message_type = GRANT;
         message.freq = getfreq(stack[2].cmd, system);
-        message.talkgroup = stack[2].full_address;
+        message.talkgroup = stack[2].address;
         if (stack[2].status >= 8) {
           message.encrypted = true;
           if ((stack[2].status == 10) || (stack[2].status == 12) || (stack[2].status == 13)) {
@@ -395,7 +395,7 @@ std::vector<TrunkMessage> SmartnetParser::parse_message(std::string s,
         //     << "[" << system->get_short_name() << "] [patch/msel termination] TG $"
         //     << std::hex << stack[3].full_address;
         message.message_type = UNKNOWN;
-        message.talkgroup = stack[3].full_address;
+        message.talkgroup = stack[3].address;
         messages.push_back(message);
         return messages;
       }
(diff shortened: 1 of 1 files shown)


ginomiglio:
Rebased with the duper-recorder changes that were just merged into master

robotastic:
Thanks! For some reason, I thought I had baked this into the duper-recorder branch... looks like I forgot.

PR #405, opened 2020-09-27
Add more info to priority logging

### LINKED ISSUE #309: Add detail to "Not recording because of priority" log entry
I am recommending / asking if more detail could be provided for the "Not recording because of priority" log entry.

It would be very handy to know:
a.   the talkgroup not being recorded
b.   the reason (blacklisted, no recorder available)

So basically, include TGID and Priority # in the log entry

ginomiglio:
@robotastic I fixed the issue [that I mentioned here](https://github.com/robotastic/trunk-recorder/commit/7c71e217da20c0ee21efa26c31dad6c047fbe891#r42754973) and added some logging as well (which addresses #309). I haven't written C++ in a long time so let me know if I did anything silly.

Everything seems to work as expected! The logging helped me verify and seems generally useful (especially since this is sort of a breaking change for people who have priorities set that will suddenly start getting enforced...)

### DIFF
--- a/trunk-recorder/main.cc
+++ b/trunk-recorder/main.cc
@@ -636,9 +636,9 @@ bool start_recorder(Call *call, TrunkMessage message, System *sys) {
 
       if (talkgroup) {
         if (talkgroup->mode == 'A') {
-          recorder = source->get_analog_recorder(talkgroup->get_priority());
+          recorder = source->get_analog_recorder(talkgroup);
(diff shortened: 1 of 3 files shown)


robotastic:
Nice - yea, I just realized I cut and pasted and gotten the available analog recorders wrong. Add the logging is a good idea and breaking it into different functions should make it clearer.

PR #413, opened 2020-10-18
Enforce MODE flag for Encrypted in `talksgroups.csv`.

### LINKED ISSUE #343: Encrypted calls are still getting recorded and uploaded && Repeat transmissions across multiple TGs
[https://openmhz.com/system/jaxp25](url)

Two part question:

A lot of encrypted calls are still getting through, any way to stop it? 

Most calls are getting stopped and it shows in the terminal:  `	Not Recording: ENCRYPTED `

But a lot are still being recorded that are obviously encrypted. hideEncrypted is set to default false.

Also I noticed with a call goes out through multiple talk groups, it records it on all, is there a way to truncate them so it does not play 4 times when ever a medical call is dispatched?

Dygear:
Fixes #343 by enforcing the `MODE` flag in the `Talkgroups.csv` file. If you set the `MODE` flag to `E` it will _always_ ignore that talk-group as encrypted.

Keep in mind that `Talkgroup::Talkgroup` mode is a single char, as such you can only set to `E` and not `DE`. The only other place this is checked is for `A` as in analog, so it shouldn't break anything else. Also fine with me as encrypted also implies that it must be digital.

### DIFF
--- a/trunk-recorder/main.cc
+++ b/trunk-recorder/main.cc
@@ -620,7 +620,7 @@ bool start_recorder(Call *call, TrunkMessage message, System *sys) {
     call->set_talkgroup_tag("-");
   }
 
-  if (call->get_encrypted() == true) {
+  if (call->get_encrypted() == true || talkgroup->mode == 'E') {
     if (sys->get_hideEncrypted() == false) {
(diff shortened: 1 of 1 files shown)


robotastic:
Great idea! Added!

PR #421, opened 2020-11-15
Plugin system

JoeGilkey:
As more and more features are added, such as uploaders and decoders, there has been a lot of code specific for each one. In order to organize this a little better, I have added a plugin system. This has the added benefit of allowing people to write their own uploaders, or custom handlers, without bloating the core trunk-recorder code. Included in this change is a basic plugin system, along with documentation to support it. Once merged in, I plan on cleaning up the uploaders to move them into this system and clean up a lot of the cross references.

Side note, this plugin system was designed to also support things like live audio streaming, and allow the plugins to only hook into things that they actually use.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -215,6 +215,8 @@ list(APPEND trunk_recorder_sources
   trunk-recorder/talkgroups.cc
   trunk-recorder/unit_tag.cc
   trunk-recorder/unit_tags.cc
+  trunk-recorder/plugins/plugin-common.cc
+  trunk-recorder/plugins/plugin-manager.cc
   lib/gr_blocks/freq_xlating_fft_filter.cc
   lib/lfsr/lfsr.cxx
   lib/gr_blocks/decoders/fsync_decode.cc
@@ -223,6 +225,7 @@ list(APPEND trunk_recorder_sources
   lib/gr_blocks/decoders/signal_decoder_sink_impl.cc
   lib/gr_blocks/decoders/tps_decoder_sink_impl.cc
   lib/gr_blocks/decoder_wrapper_impl.cc
+  lib/gr_blocks/plugin_wrapper_impl.cc
   lib/gr_blocks/selector_impl.cc
   )
 
--- a/PLUGIN-SYSTEM.md
+++ b/PLUGIN-SYSTEM.md
@@ -1,0 +1,52 @@
+Plugin System
+=============
+
+The plugin system is designed to allow third parties to integrate their own features without changing the core code of trunk-recorder, and allow for features to be updated independently of the trunk-recorder code base. The current methods supported by the plugin system is loosely based on the Status uploader.
+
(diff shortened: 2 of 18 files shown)


robotastic:
This is a great idea - I will have to go through it a bit to understand it. It would be good to also add in support for Call Status  and make websockets just one of the approaches... I will dig in tonight.

LocusAutomotive:
Thank you. I have support for call status (start and end) in there already. I was going to move the current stat_socket, call_uploader and broadcastify_uploader into using this method thru the initialize_internal_plugin call. The idea initially came about after looking into how I could add live audio streaming using a Grpc server, along with the call status.

JoeGilkey:
Thank you. I have basically cloned the stat_socket methods, so call status (start/end) is in there. It also adds support for live audio streaming. I was planning on moving stat_socket, call_uploder and broadcastify_uploaded to use this if you’re happy with it. My original intention was to add a Grpc audio streaming server, but not muck up the code for anyone not using it.

robotastic:
This looks pretty good! I have actually never done anything with dynamic libraries, so thanks for setting this up. 
I am guessing that the Audio Streamer needs access to the Recorder, so that it can get information about the Call?
It might be worth also building out hooks for Unit Registration, Affiliation and Deregistration. Someone could build a plugin to log that to a file. 

Let me go start up a branch for this.

robotastic:
I created the `plugins` branch. Let me know if that makes things more confusing though, we can just work off of this PR if that is better.

https://github.com/robotastic/trunk-recorder/tree/plugins

JoeGilkey:
The plugin branch should be fine. I’ll work off of that. I passed the recorder into the audio stream callback to make it easier to identify static (I.e. conventional) vs a dynamic (I.e. P25 talkgroup), and let the audio streaming plugin sort it out consistently. I’ll add the unit calls into it as well, that’s a good idea. I also realized I need to add a null check on the configuration loading if there is no “plugin” in the json file.

RadioTechGuy:
Excited to see what we can build on top of this.

Thanks for starting this @JoeGilkey

robotastic:
This is coming along nicely!

robotastic:
Let me know when things are looking good and I will update the branch and give it a try. I can help move over the Uploaders, if you haven't started that yet.

robotastic:
Hi @JoeGilkey I wanted to check in and see how things are going! Does this approach still look good? If so, I can try to add to it over the break.

jodfie:
This looks really exciting!

Dygear:
@JoeGilkey this looks awesome. It would be great to use in a WebRTC system so that people could follow along in their browser. As long as the TURN server was setup to handle the load, it actual computer wouldn't use that much bandwidth. It's also nice to have some real time telemetry from these systems.

robotastic:
I agree @Dygear ! Get this fully implemented and merged into the Main branch will be my next project. Once I have something stable I will let you know. I would be great if you could try making a plugin once things become stable

Issue #153, opened 2017-12-12
Per-day call logs instead of JSON files


rosecitytransit:
Since I'm not interested in uploading audio, I have never liked having call metadata in 1000's of tiny JSON files.  As I mentioned in #64 and #66, I have been using custom file names and adding the radio IDs to them and ignoring stuff like the frequency,

However, I believe a better way would be to instead offer the option of putting the metadata in a per-day call log file.  I've successfully done that on a new install and the changes I made to call.cc and call.h are here:
[rosecitytransit/trunk-recorder:kcmscanner](https://github.com/rosecitytransit/trunk-recorder/compare/master...rosecitytransit:kcmscanner#diff-fa70b60f6dffff84d1b1323b90567feb)

Two issues:
-I did it by modifying the JSON file creation code; this should instead be in addition to that and made an option in the config file.
-I'm not sure what's the best log file layout.  I used:
hhmmss,length,emergency;source(,source...);frequency,length,error,spike(;frequency,length,error,spike)
For example:
110527,34,28045,0;2800014,2803056,2800014,2803056;770.518750,0,0,0;770.268750,0,0,0;769.768750,0,0,0;770.518750,0,0,0;
with ; as a "super" delimiter dividing different sections

PR #441, opened 2021-03-13
Finished unit logging script (fixes #181, can close #440)

### LINKED ISSUE #440: Unitscript failure
I’m getting a syntax error on Line 11 (“elif”) and I’m no Bash expert - but I think elif has to have a test parameter behind it. And I don’t think “!=“ is used in Bash either?

Also the readme seems to imply that I just put the script into the ~/trunk-recorder directory, but it couldn’t find it when I ran recorder. It worked with the fully delimited name (after I first noticed the file was executable -but that’s when the syntax errors popped up.)

Last question - what’s the operational difference between this and OnUnitChange?

rosecitytransit:
I'm pretty sure this works. I have not built trunk-recorder with with the capability to call this script yet, however I did test out the script by giving it the same parameters trunk-recorder should.

Finishes #181, can close #440


rosecitytransit:
@robotastic This is working for me. You might want to test it, but I think it can be merged.

robotastic:
I compiled things and it looks good - Thanks for the changes!

rosecitytransit:
And thank you for getting to it. I went ahead and added in some other minor
changes I came up with.

My next goal is to address #153 and also add in the live Web page player
used on http://kcmscanner.ddns.net/

EllioTTDawson (commented):
#557 #368

EllioTTDawson:
gh pr checkout 441

PR #452, opened 2021-03-31
Update p25_recorder.cc

robotastic:
Moved the valve to after the down filter channelizer - This should improve performance when using sources that have large bandwidth.

### DIFF
--- a/trunk-recorder/recorders/p25_recorder.cc
+++ b/trunk-recorder/recorders/p25_recorder.cc
@@ -139,16 +139,17 @@ void p25_recorder::initialize_prefilter() {
 
 
 
-  connect(self(), 0, valve, 0);
+  
   if (double_decim) {
-    connect(valve, 0, bandpass_filter, 0);
+    connect(self(), 0, bandpass_filter, 0);
     connect(bandpass_filter, 0, mixer, 0);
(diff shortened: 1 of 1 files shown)


frameshift18:
I see higher CPU utilization after this commit. TR still works fine and audio doesn't seem to sound any different. This is running on an E3-1230v2 with a LimeSDR Mini via the SoapySDR API with 8MHz bandwidth. You can clearly see where I switched below. Also note the fairly consistent utilization after. Before it would vary with how loaded the radio system is.

frameshift18:
That was exactly my first thought actually.  Before it would usually sit around 1.6Ghz and spike up another 400Mhz or so with load/encoding.  After it sits around 2.3Ghz give or take and CPU temps are 10C higher too.  I'm going to try testing on my other system with an SDRPlay, although that is using Soapy too. Do you have any ideas how I could profile this or other suggestions?

robotastic:
Unfortunately I am not sure of a good way to profile things. I will take a better look at my system this weekend... and probably roll back this change.

leee:
@frameshift18 I know this PR is closed, but I'll be curious to see your dashboard again (freq, utilization, temp, etc.) with this change reverted.

PR #453, opened 2021-04-02
allow specifying log file directory

### LINKED ISSUE #443: Log directory should be configurable
[The log directory should be configurable](https://github.com/robotastic/trunk-recorder/blob/a6e4a8df83e89c7203fa07749f63c41cd682c2c6/trunk-recorder/main.cc#L1463).

x0ul:
This can be done with the config key "logDir". It does what you think for both relative and absolute paths. No trailing slash needed, but it won't hurt if you include one. Same behavior of a relative-path "logs" directory if none is specified.

Fixes #443, hopefully.

Tested by running it.

Tangent/newb question: Is there a dummy source/config.json available? I'm a newb and spent more time struggling to start up an RTL-SDR dongle than I did coding this up.

### DIFF
--- a/trunk-recorder/config.h
+++ b/trunk-recorder/config.h
@@ -21,6 +21,7 @@ struct Config {
   std::string instance_id;
   std::string capture_dir;
   std::string debug_recorder_address;
+  std::string log_dir;
   bool debug_recorder;
   int debug_recorder_port;
   int call_timeout;
--- a/trunk-recorder/main.cc
+++ b/trunk-recorder/main.cc
@@ -536,6 +536,8 @@ bool load_config(string config_file) {
     BOOST_LOG_TRIVIAL(info) << "Call Timeout (seconds): " << config.call_timeout;
     config.log_file = pt.get<bool>("logFile", false);
     BOOST_LOG_TRIVIAL(info) << "Log to File: " << config.log_file;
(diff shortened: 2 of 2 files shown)


bctrainers:
> Tangent/newb question: Is there a dummy source/config.json available? I'm a newb and spent more time struggling to start up an RTL-SDR dongle than I did coding this up.

The Github TrunkRecorder wiki has a couple of examples posted. If someone else hasn't done so already, they may need to be slightly tweaked/updated to support the v2 stuff.

x0ul:
Ah, what I meant by "dummy source" was a virtual device so that actual SDR hardware isn't needed in order to start the program. That would have made it a little easier to test.

I tried a few example configs and found `examples/config-park-conv.json` got me started up :)

rosecitytransit:
In theory, you could run it against the file of a saved raw signal. I'm pretty sure GNURadio, which is what trunk-recorder uses underneath, supports this.

robotastic:
Genius! Thanks for the addition. 

It would definitely be possible to add in support for reading from a file... which would make testing a lot easier. GnuRadio supports file sources and you can dump from a program like GQRX. There may even be a hacky way of doing now with the gr-osmosdr drivers: https://osmocom.org/projects/gr-osmosdr/wiki#IQ-File-Source

PR #458, opened 2021-04-25
Add low pass filter to analog recorder audio flow

leee:
Analog FM land mobile radio has an audio frequency response range of 300 to 3000 Hz [0]. In this change, we add a low pass filter to make our analog recorder consistent with that fact.

We characterize the new low pass filter with the "cutoff" (center) at 3250 Hz, and a transition band of 500 Hz. This produces a final audio signal with a pass band of 300 to 3000 Hz (including the already included `high_f` filter), a transition band of 3000 to 3500 Hz, and a stop band beyond that. We could use a band pass filter targeting 300 to 3000 Hz, but desire a more gradual rolloff for our low pass instead of an aggressive brick wall at both ends.

Tested working on top of `robotastic/master@b0e96e4` overnight.

Some before/afters below. Transmissions from systems that already sounded good at this site sound even better without the hiss and the harsh harmonics in alert tone 1:

 | 
:-:|:-:
*fig. 0. FFT of [this audio](https://gist.githubusercontent.com/leee/1807c39b3afd5f70a709b5711ba7be05/raw/bdb607cbfab68f2c8c6f418cea1ca34637031911/fdnyems-1-1619342789.m4a) on `robotastic/master@b0e96e4`.* | *fig. 1. FFT of [this audio](https://gist.githubusercontent.com/leee/1807c39b3afd5f70a709b5711ba7be05/raw/bdb607cbfab68f2c8c6f418cea1ca34637031911/fdnyems-1-1619342788.m4a) with the proposed change.*

And an example of higher sample rate audio not helping at all. What's the use when it's all noise, coming from equipment and systems designed for audio below 4 kHz?

 | 
:-:|:-:
*fig. 2. FFT of [this audio](https://gist.githubusercontent.com/leee/1807c39b3afd5f70a709b5711ba7be05/raw/bdb607cbfab68f2c8c6f418cea1ca34637031911/nypd-1-1619324653.m4a), exhibiting some annoying noise, not as bad as the one we were originally investigating!* | *fig. 3. FFT of [this audio](https://gist.githubusercontent.com/leee/1807c39b3afd5f70a709b5711ba7be05/raw/bdb607cbfab68f2c8c6f418cea1ca34637031911/nypd-1-1619324654.m4a) simultaneously recorded [1]. We don't need any of that noise beyond 3-4 kHz!*

[0] - an artifact of copper telephony, this is well displayed in the specification and use of usable voice frequency band for radio remote control (e.g. tone remote signaling).
[1] - different test server, but at the same receive site, off of the same receive multicoupler, and identical SDR parameters.

### DIFF
--- a/trunk-recorder/recorders/analog_recorder.cc
+++ b/trunk-recorder/recorders/analog_recorder.cc
@@ -135,7 +135,7 @@ analog_recorder::analog_recorder(Source *src)
   squelch_two = gr::analog::pwr_squelch_ff::make(-200, 0.01, 0, true);
 
   // k = quad_rate/(2*math.pi*max_dev) = 48k / (6.283185*5000) = 1.527
-  
+
   int d_max_dev = 5000;
   /* demodulator gain */
   quad_gain = system_channel_rate / (2.0 * M_PI * d_max_dev);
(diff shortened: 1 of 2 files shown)


robotastic:
This was merged in, along with some additional analog related code in PR #467

leee:
@robotastic sounds great, but none of the commits on `main` (`master`) since `b0e96e4` don't touch the analog recorder:

```
$ git rev-parse --short HEAD
dc90ae4

$ git --no-pager diff b0e96e4 HEAD --name-only
Dockerfile
blacklist-rtl.conf
trunk-recorder/call.cc
trunk-recorder/call.h
trunk-recorder/call_conventional.cc
trunk-recorder/call_conventional.h
```

I think you may have tested this but forgot to mark it for inclusion in the latest batch of work?

robotastic:
Yep - I thought I had based the analog-test branch off the PR... but I guess not. Merging this PR.

leee:
I remember you did just that, but then I think when you worked on call duration/timers, you started some work, asked for some testing, then cut it all and started fresh off of main/master but used the same branch name or something and just didn't add it back after 🙂

PR #468, opened 2021-05-15
Removes the older v2 audio player.

Dygear:
This simply removes the older V2 audio player that's actually older than the audioplayer.php code base. This PR is also regarding on if this is an appropriate action.


Dygear:
The `lib/op25_repeater/lib/*.h linguist-language=C++` change to `.gitattributes` ensure that files are correctly understood as C++ code.

rosecitytransit:
I'd like to note that, originally, v2 was my newer (well, later) version, but then others decided to modify the non-v2 version

robotastic:
Thanks for helping to clear this up! I was confused on the lineage. I also had no idea about the .gitattributes. I will have to look into that more.

Issue #456, opened 2021-04-05
Issues I'd like to (see) address(ed) (updated 5/23)


rosecitytransit:
Some of these I do want to work on and can submit myself, and this is somewhat a note for myself. Others are kind of wish list items or ideas. I've mentioned many of them on Gitter chat.

**MERGED AND GOOD**
- [x] perhaps 0eb05cd (is this an issue for anyone? Need to get other P25 systems set up and see if/how they do unit-unit calls)
- [x] decoding rate chart creator ( f36488f or eb9a000 but also want non-PHP shell script version )
- [x] make log files match console output, include message severity levels (I've done this)
- [x] make Decim messages go to BOOST logging instead of console output https://github.com/robotastic/trunk-recorder/blob/master/trunk-recorder/recorders/p25_recorder.cc#L77 https://github.com/robotastic/trunk-recorder/blob/master/trunk-recorder/recorders/debug_recorder.cc#L61 https://github.com/robotastic/trunk-recorder/blob/master/trunk-recorder/systems/smartnet_trunking.cc#L66 https://github.com/robotastic/trunk-recorder/blob/master/trunk-recorder/systems/p25_trunking.cc#L60
- [x] cleanly exit by closing open calls (including writing JSON files)

**MERGED BUT MAY HAVE ISSUES**
- [x] controlWarnRate message frequency (sometimes I want it every 3 seconds when adjusting antennas, other times 1/min to make sure things are good)
*(see note on commit 2d6a954 and discussion at https://github.com/robotastic/trunk-recorder/pull/470#discussion_r636501481)*
- [x] silenceFrames (3b3fa38) I'm guessing this should really be a boolean like it originally was, that it should do "real time" recording of calls, as in including the silence between transmissions
- [x] detect P25 call terminations and close calls based on them (#451, plan on working on at rosecitytransit/trunk-recorder:newtrimetscanner)
*(see comment at https://github.com/robotastic/trunk-recorder/pull/480#issuecomment-846443248)*
- [x] log call priority and phase 2 slot #
*(priority, and duplex/mode, logging added w/67a0d86, but may not be complete; also slot logging not done)*

**MAYBE OTHERS CAN DO**
- [ ] debugRecorder documentation, both for SDR# and Wireshark (both control channel and calls) use *([Started](https://github.com/robotastic/trunk-recorder/wiki/Using-Debug-Recorders))*
[...]

Issue #200, opened 2018-09-10
Need help upgrading, not getting P25 radio IDs


rosecitytransit:
I am trying to update from the ancient gr-latency branch from which @robotastic added unit-unit call support and am having some issues.  It appears that in this commit https://github.com/robotastic/trunk-recorder/commit/3f64de33e23408f95912826ff7d5b827919ac9f5 the code was changed from getting radio IDs from the voice packet metadata to getting them from the control channel.  However, the system I record only occasionally puts radio IDs on the control channel.

I use radio IDs to handle unit-unit calls which don't show up correctly on the control channel so it is important to have them.  I am wondering if someone can change my https://github.com/rosecitytransit/trunk-recorder/tree/trimetscanner branch so it gets radio IDs from voice packets.

In addition, I am trying to start recording from a second site, and at first things work great but after half a day or so, I start getting an increasing amount of 44 byte files.  I don't think it's a signal issue as the decoding issue remains great.  I am hoping the upgrade will address this.

kg6uyz:
I was having the 44 byte file problem.  Make sure you have a tuned antenna
for the frequency spectrum your recording..  I also bought some big heat
sinks to dissipate heat on my sdr's that ran super hot.  Seems to have made
things better....

rosecitytransit:
I'm not sure that's the issue as I'm using the same bandwidth settings when I add the second site (1.8 MHz for each SDR) and I thought all of the work was done in software, meaning the same data is getting pulled in.

Also, if there's not an antenna covering a frequency, no file will show up since it won't even try to record.

rosecitytransit:
I still haven't solved this problem, and a listener is willing to pay whoever to help solve it.  I'd really like to figure out why audio starts failing to record with the second site, but if I could at least get the program updated that would be a success.

rosecitytransit:
Still interested in compensating someone to help me with this

rosecitytransit:
Well, I don't know if they're coming from the control channel or the voice frames, but I appear to be getting the radio IDs.

I'm now trying to detect call terminations on the voice channels. So far, I have 871f6a3 but I get segmentation faults every time a call ends.

PR #470, opened 2021-05-16
various enhancements (closes most of #456, closes #451)

### LINKED ISSUE #451: Closing calls based on voice channel termination
On one of the systems I monitor, update messages for unit-to-unit calls (which is where the juiciest action is) do not regularly appear on the control channel like it seems they should.

For the old version of trunk-recorder I'm using, as noted in #64 I modified the code to set the `call_timeout` very high just for those calls and it's been working mostly well. Portions of some calls do get missed, some get split up into multiple files, and occasionally a part of the next call gets grafted on to the end.

However, I'd like to instead close calls when either a termination message is seen on the voice channel or when the stream of messages there ends. @robotastic has noted that something similar can't be done for analog calls since it's not possible to see the signal levels, but I thought that for digital calls it could be done in conjunction with, or in place of, the existing `call_timeout` logic.

I noticed in the many_files and analog_squelch branches (e.g. this commit c7cd9a1 ) some work with voice call terminations and am trying to get them to work.  So far I have 871f6a3 but I get Segmentation Faults when a call closes. Code is based on attempting to mimic `rx_status` flow.

Right now, it resets the `call_timeout` if the call hasn't ended, but it would be less intensive if instead the p25 framer could quit/`return` when the call has terminated.

rosecitytransit:
See list I worked from at #456 (also closes #451)

Notes:
- I'm guessing silenceFrames should really be a boolean like it originally was, that it should do "real time" recording of calls, as in including the silence between transmissions
- I did not include logging of P25 phase 2 slot # like I was wanting
- I'd really like to have a daily counter in `Call::create_filename()` that can be put in the daily log file/on the live Web page; it would allow calls to have a fixed row number based on start time (even if an earlier-started call ends/arrives later) and identify when calls are not listed on the Web page
- I'd like to have radio IDs from the control channel and voice frames logged to 2 separate lists
- I'm wondering if there's a better way to do the daily log rather than opening and closing the file each call, such as using Boost Log
- call priority, duplex, circuit/packet mode may not have been implemented in all places
- I re-did all of the commits (by saving them as patches before hard resetting); hopefully everything got put back


lib/gr_blocks/nonstop_wavfile_sink_impl.cc
```diff
@@ -267,7 +267,7 @@ int nonstop_wavfile_sink_impl::dowork(int noutput_items,  gr_vector_const_void_s
   //long curr_src_id = 0;
 
   for (unsigned int i = 0; i < tags.size(); i++) {
-    if (pmt::eq(this_key, tags[i].key) && d_current_call->get_system_type() == "conventionalP25") {
+    if (pmt::eq(this_key, tags[i].key) ) {
```
robotastic, line 270: I think this could lead to trouble. Sometimes the Control Channel is slightly ahead of the decoding on the Voice Channel, so there maybe a lot of oscillation between source IDs. Is there a particular reason for making this change? Are source IDs different between the 2 channels?

trunk-recorder/main.cc
```diff
@@ -659,9 +663,9 @@ bool start_recorder(Call *call, TrunkMessage message, System *sys) {
       //int total_recorders = get_total_recorders();
 
       if (recorder) {
-        if (message.meta.length()) {
-          BOOST_LOG_TRIVIAL(trace) << message.meta;
-        }
+        /* if (message.meta.length()) {
```
robotastic, line 666: I am confused why this section was commented out. If it is too verbose for normal use, lets leave it as TRACE vs DEBUG

trunk-recorder/main.cc
```diff
@@ -893,6 +897,14 @@ void unit_ack(string unit_script, string shortName, long unit) {
   }
 }
 
+void unit_locreg(string unit_script, string shortName, long unit, long talkgroup) {
```
robotastic, line 900: What does this function do? It looks like it mostly calls an external script

trunk-recorder/main.cc
```diff
@@ -1026,7 +1042,11 @@ void handle_message(std::vector<TrunkMessage> messages, System *sys) {
     switch (message.message_type) {
     case GRANT:
     case UPDATE:
-      handle_call(message, sys);
+      if ((message.talkgroup != 0) && (message.freq != 0)) {
+        handle_call(message, sys);
+      } else {
+        BOOST_LOG_TRIVIAL(trace) << "Skipping call: Freq " << message.freq << " TG " << message.talkgroup << " Src " << message.source;
```
robotastic, line 1048: Why would talkgroup or the Freq be 0? If this is a bad message, lets handle it further up the processing stream, in the parser.

trunk-recorder/main.cc
```diff
@@ -1056,6 +1076,10 @@ void handle_message(std::vector<TrunkMessage> messages, System *sys) {
       unit_ack(sys->get_unit_script(), sys->get_short_name(), message.source);
       break;
 
+    case LOCREG:
```
robotastic, line 1079: I will probably see this defined latter on.... but I am confused what LOCREG stands for.

trunk-recorder/main.cc
```diff
@@ -1197,7 +1221,20 @@ void monitor_messages() {
   while (1) {
 
     if (exit_flag) { // my action when signal set it 1
-      printf("\n Signal caught!\n");
+      BOOST_LOG_TRIVIAL(info) << "Signal caught! Closing calls...";
```
robotastic, line 1224: Nice - this is a good idea. It looks like this should clean up things and shutdown all the open calls.

robotastic:
Thanks @rosecitytransit !! This looks like it should clean up the silence/idle frame stuff that had been lurking in the code. 

I am a little hesitant to add in the **daily-log** functionality. I am trying to keep the core of TR focused. Would it be possible to something similar as part of the upload-script? I would process the associated .json file and append a line to the end of the log file. This seems like it would be a great plugin. I am going to focus on getting the plugin system working next. It would allow for these types of customization. Can you separate out the **daily-log** changes as a separate PR?

One thing I will have to thoroughly double check in the future, is how TR and OpenMHZ handle timezones and the concept of a day. TR uses local time I think and openmhz uses that time... I think. 

Can you tell me a little more about Voice Chan vs Control Chan source IDs? Do they differ ever? Would there be a reason for tracking them separately?

For future PRs, it is easiest if you keep them focused on a single set of changes. It makes it easier to see what parts are related... and I can also tackle small ones quicker. :) 

Thanks for trying to make TR better!!

rosecitytransit:
I do apologize that I waited to submit everything at once. I know I should have done it as I went along. As can be seen [here](https://github.com/rosecitytransit/trunk-recorder/commits/master?after=4efde123e678d130698b96b77e165da87d58d752+0&branch=master) or [here](https://github.com/rosecitytransit/trunk-recorder/commits/newtrimetscanner), I had some of it done 25 days ago.

I am willing to reset and rebuild the commits again (I have the patches created/saved) and either leave out the unwanted ones, or just submit them piece-by-piece.

lib/gr_blocks/nonstop_wavfile_sink_impl.cc
rosecitytransit: This addressed #200 for me, with IDs not showing on the control channel. If no one else is experiencing the problem, this can be skipped. As I noted, what I'd like to do is log the control channel and voice frame sources separately (I'm envisioning two separate srcLists in the JSON files).

trunk-recorder/main.cc
rosecitytransit: As shown in the commit https://github.com/robotastic/trunk-recorder/commit/917abd9f3fc427c966cb9b79670923df4a2c395b I'm sending messages to the log in p25_parser.cc, as in before returning to main.cc and this line. This allows different messages to be logged at different severity levels.

trunk-recorder/main.cc
rosecitytransit: Yes, it calls the unitScript I added in #181/#441 on Location Registration Response messages (I noticed that I was getting them, and saw in the standards that they contained radio and talkgroup IDs).

trunk-recorder/main.cc
rosecitytransit: I've consistently (as in continuously) seen messages with them. I want to say they even show up in UniTrunker. I'm thinking it's not due to poor signal. I could send you some raw signal if you want to see it.

trunk-recorder/main.cc
rosecitytransit: Location Registration Response. I (we?) should maybe use official P25 opcode aliases, in this case `LOC_REG_RESP'

trunk-recorder/main.cc
rosecitytransit: Yeah, I hated ending up with orphan WAV files (unconverted and with no corresponding JSON file).

lib/gr_blocks/nonstop_wavfile_sink_impl.cc
robotastic: I want to dig into this more before adding in this change for everybody. Taking in Source IDs for a recording in 2 different spots seems like trouble. Could it be that some of the control channel messages are being processed incorrectly? Another alternative would be to have all of the Source IDs come from the Voice Channel. This would probably make the Source IDs line up better with the recordings.

Let's leave this change out of the PR and add to items to the list:
 - investigate why no Source IDs are in the Control Channel Messages
 - Look to switching all Source IDs to come from the Voice Channel.

trunk-recorder/main.cc
robotastic: Gotcha - This could be another processing error. I have never see this in my system. The error checking as part of the decoding should stop malformed packets because of low signal strength but maybe some get through. Can you turn on Tracing level debug so we can see exactly which message parsing function is causing this? This error check should be integrated directly into the parsing.

Could you also a Radio Reference link for this system? I am curious if there could be some special manufacturer messages in the control channel.

Let's leave this out of the PR and handle it separately.

trunk-recorder/main.cc
robotastic: This is a great feature - lets keep it in the PR

trunk-recorder/main.cc
robotastic: I am fine with the shorter version, 'LOCREG'. Can you add in comments for the function and the switch statement?

robotastic: Let's keep this in the PR

trunk-recorder/main.cc
robotastic: Gotcha - Can you delete the if statement instead of commenting it?

trunk-recorder/main.cc
robotastic: Gotcha - Let's leave this in? Can you also add short comment in on what the type of message is and what the function does?

robotastic:
Thanks @rosecitytransit !  I went through and tried to slice it up in to changes that could be approved right away vs others that need more research. 

I want to to try and figure out why the Source IDs are not in the Control Channel but are in the Voice Channel. 

I like the daily log concept, but I would like to find a more flexible way of doing it. People may want different time scales of summerization or for it to end up somewhere besides a text file. Having it be part of the Upload Script would give that flexibility, but ultimately it would be a great use of the Plug-In system... I will go focus on getting that working.

Can you go up the PR with the features that I tagged up above to include?

trunk-recorder/main.cc
rosecitytransit: I don't see the problem with it, but I can definitely leave it out. As I said, I think it's not a result of malformed packets.

Your mention of trace debug level reminded me of how trunk-recorder can show the raw data bits (which would include packet check sums). I should be able to ferret out the ones in question and copy them here. [This is the system I primarily monitor](https://www.radioreference.com/apps/db/?sid=7202); I have at least 2 other P25 systems I can check too.

trunk-recorder/main.cc
rosecitytransit: Will do.

lib/gr_blocks/nonstop_wavfile_sink_impl.cc
rosecitytransit: As I said in my original comment, "I'd like to have radio IDs from the control channel and voice frames logged to 2 separate lists". Given that this was really just a quick fix for me, I'm definitely willing to leave it out.

[This system](https://openmhz.com/system/osrppdxc) is also using this code, but so much of the calls are patched from elsewhere.

rosecitytransit:
To clarify before I rebuild things:
modify main.cc in https://github.com/robotastic/trunk-recorder/pull/470/commits/917abd9f3fc427c966cb9b79670923df4a2c395b

leave out
https://github.com/robotastic/trunk-recorder/pull/470/commits/b69d64de795e57b87ba9ae3d3b6c76b0baa791ad

modify https://github.com/robotastic/trunk-recorder/pull/470/commits/7a3447b22845d8851fc779ddd36103bfb81f5249 to remove daily log parts (and maybe put controlWarnUpdate README entry with it's commit)

leave out https://github.com/robotastic/trunk-recorder/pull/470/commits/1bd55e1ab485660eaef4629e0599de973ed99ef6

leave out and look into https://github.com/robotastic/trunk-recorder/pull/470/commits/9792d9525faf44ed3006f59e1ae66f6b47188c04

leave out https://github.com/robotastic/trunk-recorder/pull/470/commits/1bd55e1ab485660eaef4629e0599de973ed99ef6

leave out https://github.com/robotastic/trunk-recorder/pull/470/commits/d210033835be32797b0be9e2748c7a89dbb1c75b since it uses the daily log

also what's your thought on https://github.com/robotastic/trunk-recorder/pull/470/commits/21085192bfb1b105fa7bc089d86acceddfcc0df6 ? I only really need it due to lack of control channel updates for unit-unit calls, and besides that it's not good for transmission mode (nor is the re-tune function).

robotastic:
That is much clearer - I didn't realize they were sort of separated by Commit. 

I miseed https://github.com/robotastic/trunk-recorder/commit/21085192bfb1b105fa7bc089d86acceddfcc0df6 - lets add that into the PR. It seems like it would only add functionality. 

It is probably worth reinvestigating how Errors are calculated and recorded. I think there is some better code that was added to OP25 that could be used in the future.

What you propose above looks good to me. Can you submit a PR based upon that?

rosecitytransit:
OK, I reset my branch and (hopefully correctly) re-added just the commits you were OK with. It should be ready to go now.

trunk-recorder/systems/p25_parser.cc
```diff
@@ -421,6 +453,9 @@ std::vector<TrunkMessage> P25Parser::decode_tsbk(boost::dynamic_bitset<> &tsbk,
     message.source = sa;
     message.emergency = emergency;
     message.encrypted = encrypted;
+    message.duplex = duplex;
+    message.mode = mode;
+    message.priority = priority;
```
leee, line 458: From the [buildx @ 86689ce](https://github.com/robotastic/trunk-recorder/runs/2624228538#step:7:3420):
```
#13 300.5 /src/trunk-recorder/systems/p25_parser.cc:456:22: error: 'duplex' was not declared in this scope
#13 300.5   456 |     message.duplex = duplex;
#13 300.5       |                      ^~~~~~
#13 300.5 /src/trunk-recorder/systems/p25_parser.cc:457:20: error: 'mode' was not declared in this scope; did you mean 'modf'?
#13 300.5   457 |     message.mode = mode;
#13 300.5       |                    ^~~~
#13 300.5       |                    modf
#13 300.5 /src/trunk-recorder/systems/p25_parser.cc:458:24: error: 'priority' was not declared in this scope
#13 300.5   458 |     message.priority = priority;
#13 300.5       |                        ^~~~~~~~
```

@rosecitytransit do [these lines](https://github.com/robotastic/trunk-recorder/pull/470/files#diff-f75ab21e8810348f73654b768283e40a0d9948d9a1d89c12cd73fd55c5b6cba4R442-R444) need to be uncommented out?

rosecitytransit:
Looks like patch failure got saved in the code. Thanks for noticing it and
I'll look into it tonight.

trunk-recorder/systems/p25_parser.cc
rosecitytransit: Yep, they did. I was thinking it should have been done in the patch I applied in rebuilding to create this PR, but it looks like I may have never have changed those lines.

rosecitytransit:
should be good to go

trunk-recorder/systems/p25_parser.cc
```diff
@@ -562,13 +609,13 @@ std::vector<TrunkMessage> P25Parser::decode_tsbk(boost::dynamic_bitset<> &tsbk,
     message.message_type = DEREGISTRATION;
     message.source = si;
 
-    BOOST_LOG_TRIVIAL(debug) << "tsbk2f\tUnit Deregistration ACK\tSource ID: " << std::setw(7) << si << std::endl;
+    BOOST_LOG_TRIVIAL(debug) << "tsbk2f\tUnit Deregistration ACK\tSource ID: " << std::setw(7) << si;
   } else if (opcode == 0x30) { //
-    BOOST_LOG_TRIVIAL(debug) << "tsbk30";
+    BOOST_LOG_TRIVIAL(trace) << "tsbk30 TDMA SYNCHRONIZATION BROADCAST";
```
leee, line 614: trace or debug?

trunk-recorder/main.cc
```diff
@@ -539,7 +539,9 @@ bool load_config(string config_file) {
     config.log_dir = pt.get<std::string>("logDir", "logs");
     BOOST_LOG_TRIVIAL(info) << "Log Directory: " << config.log_dir;
     config.control_message_warn_rate = pt.get<int>("controlWarnRate", 10);
-    BOOST_LOG_TRIVIAL(info) << "Control channel warning rate: " << config.control_message_warn_rate;
+    BOOST_LOG_TRIVIAL(info) << "Control channel decoding warnings (< messages/second): " << config.control_message_warn_rate;
+    config.control_message_warn_updates = pt.get<float>("controlWarnUpdate", 3.0);
```
leee, line 543: In the property tree, you get `controlWarnUpdate` as a float, but don't you describe it as an int in seconds?

rosecitytransit:
I believe it should be ʼtraceʼ because it is a repetitive system message.
I think i changed it setting up a system where I saw it.

rosecitytransit:
I was wondering about that and asked in the commit message "should it (and
timeDiff and statusTimeDiff) be ints instead of floats?"

robotastic:
Good point on the Time Diff-ing being an int, vs a float. Not sure why I did it that way. Let's merge what you have in. We can can move those calculations over to be ints down the road.

trunk-recorder/systems/p25_parser.cc
rosecitytransit: I believe it should be ʼtraceʼ because it is a repetitive system message.
I think i changed it setting up a system where I saw it.

PR #472, opened 2021-05-16
Treat Broadcastify duplicate skips as a success

### LINKED ISSUE #361: Broadcastify Calls - Duplicate Call Handling
Broadcastify handles duplicate calls by generating an error (Broadcastify Metadata Upload Error: SKIPPED---ALREADY-RECEIVED-THIS-CALL). Unfortunately, this error seems to prevent Trunk Recorder from deleting the source audio .wav, the converted m4a, or the metadata.

In a setup where talkgroups are sometimes simulcast, and sometimes not, it's necessary to have the same talkgroups across multiple systems which makes this a fairly common occurrence.

It would be very nice if Trunk Recorder could consider a "SKIPPED" error as a success.

leee:
Resolves #361

@Ares4570 tested.


robotastic:
Excellent!

PR #480, opened 2021-05-20
Add maximum call duration to split long calls

### LINKED ISSUE #197: Enhancement: Implement a Mechanism to End Recording If A User-Defined Length is Exceeded

leee:
Resolves #197, resolves #357.

Co-authored-by: jarek <jarek@citizen.com>
Co-authored-by: jareklupinski <jareklupinski@users.noreply.github.com>
Co-authored-by: Eric Tendian <erictendian@gmail.com>

---

I'd like to reopen for consideration #357, which @EricTendian submitted on behalf of @jareklupinski. #357 directly resolves #197, although more work can be done surrounding the topic.

There are many cases where a conventional or trunked resource stays in use for extremely long periods of times, making the receipt of that one transmission delayed for the duration of that period.

#197 is an example of dispatch consoles being able to mark a trunked resource in emergency, reserving and holding voice talkpaths for it for the duration of the emergency.

Some conventional channels can experience lengthy recordings due to emergencies or high acuity incidents, and never gets a chance to meet `callTimeout` (default 3s). Some conventional systems can also be configured to have extremely long drop out delay, or stay keyed until released by console.

Ideally, calls would be terminated by transmission immediately instead of waiting for a timeout - so by carrier/power squelch, PL/DPL tone squelch, voice termination data, or control channel data. There is already some discussion on this underway in #451 which covers P25 TDU/TDULCs, and I think that issue could be generalized to also cover call termination from control channel data and conventional channels.

There's also another matter of audio files being shortened because squelched periods are not written to files, and therefore not real-time. If trunk-recorder changed its transmission strategy to individual transmissions this wouldn't be a problem, and could make transmission dedupe from multiple site recorders easier to handle, as touched on in #326.

---

I tried to merge #357 but there's been a lot of work since then. I did cherry pick and edit the pull request commit, but basically had to reorganize and rewrite it.

With all that said, this PR implements #197 - if the length of call audio exceeds a user-specified duration, it will end the call and start a new one where it left off. If no duration is specified it will default to 0 (the current behavior). To clear up the confusion from #197, the duration is measured by the length of active transmission/audio, not real-time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,6 +144,7 @@ Here are the different arguments:
  - **defaultMode** - Default mode to use when a talkgroups is not listed in the **talkgroupsFile**. The options are *digital* or *analog*. The default is *digital*. This argument is global and not system-specific, and only affects Type II `smartnet` trunking systems which can have both analog and digital talkpaths whereas `p25` trunking systems don't have analog talkpaths.
  - **captureDir** - the complete path to the directory where recordings should be saved.
  - **callTimeout** - a Call will stop recording and save if it has not received anything on the control channel, after this many seconds. The default is 3.
+ - **maxDuration** - If a call is being recorded and the duration exceeds this value in seconds, the call will stop recording and a new call will pick up where the previous on left off. The default is *0* or "no timeout".
  - **logFile** - save the console output to a file. The options are *true* or *false*, without quotes. The default is *false*.
  - **frequencyFormat** - the display format for frequencies to display in the console and log file. The options are *exp*, *mhz* & *hz*. The default is *exp*.
  - **controlWarnRate** - Log the control channel decode rate when it falls bellow this threshold. The default is *10*. The value of *-1* will always log the decode rate.
--- a/README.md
+++ b/README.md
@@ -87,7 +87,7 @@ This file is used to configure how Trunk Recorder is setup. It defines the SDRs 
 }
 ```
 Here are the different arguments:
- - **ver** - the version of formatting for the config file. **This should be set to 2**. Trunk Recorder will not start without this set. 
+ - **ver** - the version of formatting for the config file. **This should be set to 2**. Trunk Recorder will not start without this set.
  - **sources** - an array of JSON objects that define the different SDRs available. The following options are used to configure each Source:
    - **center** - the center frequency in Hz to tune the SDR to
    - **rate** - the sampling rate to set the SDR to, in samples / second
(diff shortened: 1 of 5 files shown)


robotastic:
Thanks for reviving this. It looks good to me and would address the issues outlined above, where a recording gets to be very long. 

More conversation will be need for exactly how to approach Call Termination in Trunk Systems. The `many_files` branch handles this. I will merging that in shortly. One of the challenges is that TDU messages appear on the voice channel and not the control channel. Those message end up at the wav_sink(), which is running on a separate thread. The paradigm it uses is: treating a Call like a channel reservation and then adds the concept of transmissions for each unique broadcast for a source.

Anyhow - this looks good. Merging it in! We may need to re-imagine how recordings are managed in the future to holistically cover squelch, tones and trunked systems.

rosecitytransit:
I see this is already merged, but I'd like to note that in #470 with commits https://github.com/robotastic/trunk-recorder/commit/366cd3bae8acdeba973b8ca1c9afc11fc7e78f6f and https://github.com/robotastic/trunk-recorder/commit/86689cefe2d81d39cbf6736573eaa07558028355 calls are kept open until all traffic (including post-voice TDU/TDULC messages) on the voice channel is finished. I spent a ton of time trying to use the `call_termination` I saw in the `many_files` branch but couldn't get it to work (I wish I knew a C++ programmer I could pay to help).

I needed that on the primary system I record due to unit-unit calls not having continual messages on the control channel like they should and a desire to keep parts of them together, but I know it goes against the transmission mode idea and may not be desired in general.  The re-tune feature (https://github.com/robotastic/trunk-recorder/blob/master/trunk-recorder/main.cc#L968), also goes against transmission mode, but that's a different issue.

This diff https://github.com/robotastic/trunk-recorder/commit/366cd3bae8acdeba973b8ca1c9afc11fc7e78f6f#diff-0fd516fd4624adf8f6dcf072301d849ada9e3a208fb6ef924c4a6ede842e2902 shows the statements that could be changed to instead consider calls done when the last voice frame is seen.

Also, I have seen on rare occasions a channel/frequency gets held open for many hours. While technically one "conversation" assuming post-voice frames are counted or messages continue on the control channel, I agree it gets unwieldy to have such a long file. But my preference would be to do a "soft" cut, as in close the call the next time it enters the `call_timeout` period. Though I don't know how conventional calls work, and if they'd enter that.

PR #494, opened 2021-07-17
Adjust analog_recorder deemphasis filter

### LINKED ISSUE #478: LMR FM pre-emphasis uses a time constant of 750 µs
https://github.com/robotastic/trunk-recorder/blob/915143caa674176bad26ecc0f3057aa3e71309b6/trunk-recorder/recorders/analog_recorder.cc#L148

Right now, we use a tau of 75 µs in our de-emphasis circuit because it was what was in [gnuradio/gr-analog/python/analog/fm_emph.py](https://github.com/gnuradio/gnuradio/blob/91ccd28abae6727509e88ff60032057483ac1b90/gr-analog/python/analog/fm_emph.py):

https://github.com/robotastic/trunk-recorder/blob/915143caa674176bad26ecc0f3057aa3e71309b6/trunk-recorder/recorders/analog_recorder.cc#L19

But 75 µs is meant for BCB FM in the US, and 50 µs in the EU, [as commented in that same file](https://github.com/gnuradio/gnuradio/blob/91ccd28abae6727509e88ff60032057483ac1b90/gr-analog/python/analog/fm_emph.py#L22).

```
        tau: Time constant in seconds (75us in US, 50us in EUR) (float)
```

A 75 µs RC time constant gives you a cutoff frequency of ~2122 Hz. This makes sense of course for the 30 Hz to - 15 kHz audio bandwidth of wideband BCB FM audio. 50 µs also gives you a `f_c` of ~3183 Hz, applicable in the EU, as commented in the original file.

But the original example doesn't clarify that for LMR FM, a time constant of 750 µs is used - this results in a `f_c` of ~212 Hz.

Right now, since trunk-recorder uses a time constant of 75 µs, the -6 dB/octave de-emphasis filter effectively only gets applied on audio starting at 2122 Hz to the stop band of LMR FM audio at ~3000-3500 Hz. This has the unintended result of keeping audio pre-emphasized, with a rolloff at the end that sort of looks like an allowed lower control limit rolloff as specified in the EIA emphasis spec :-)

The easy thing would be to just change tau to 750 µs. But this unfortunately has the downside of making audio sound weaker and less bright which users of trunk-recorder might have gotten used to.

With our recent addition of a low pass filter, we're already messing with the sound of things losing imagined sibilance. It might be nice to leave things alone, or it might be desirable to actually "fix" this, boost gain at the end, and consider sound processing techniques in the future.

I'd like to keep this issue open for a little bit to see what people think - if there's not much or no pushback, I'll submit a PR to change tau to 750 µs and actually de-emphasis analog audio as originally intended.

leee:
Use a time constant of 750µs instead of 75µs.

Resolves #478.

### DIFF
--- a/trunk-recorder/recorders/analog_recorder.cc
+++ b/trunk-recorder/recorders/analog_recorder.cc
@@ -145,7 +145,7 @@ analog_recorder::analog_recorder(Source *src)
   valve->set_enabled(false);
 
   /* de-emphasis */
-  d_tau = 0.000075; // 75us
+  d_tau = 0.00075; // 750us
   d_fftaps.resize(2);
   d_fbtaps.resize(2);
   calculate_iir_taps(d_tau);


leee:
This build failed here: https://github.com/robotastic/trunk-recorder/runs/3092690444?check_suite_focus=true#step:7:13809

@robotastic this PR is still safe to merge.

@robbiet480 would you be willing to take a look at this?

robotastic:
Sounds good - lets push this and if people complain we can look at making it configurable.

PR #513, opened 2021-09-13
Adds docker-client To Dockerfile for DID Support

MaxwellDPS:
Adds the docker client so those who run trunk-player/other webUI's in docker can call the host docker daemon with  in the upload script 

Example:
```bash
docker -v /var/run/docker.sock:/var/run/docker.sock trunkplayer_app_1 'python ./mange.py add_transmission $1'
```


robotastic:
That seems cool - merging in!

PR #518, opened 2021-09-18
Dockerfile build/install AUR package

toddejohnson:
This provides a Dockerfile that will build/install the AUR package: https://aur.archlinux.org/packages/trunk-recorder-git

This could easily be changed by VAR to also build: https://aur.archlinux.org/packages/trunk-recorder

I hope this can help troubleshooting the segfault I've been experiencing. Perhaps be a first step to packaging.

### DIFF
--- a/utils/Dockerfile.arch-latest-aur.dev
+++ b/utils/Dockerfile.arch-latest-aur.dev
@@ -1,0 +1,21 @@
+FROM archlinux:latest AS base
+
+# Install enought to run makepkg 
+# PKGBUILD from the repo will pull all dependencies for build and install
+RUN pacman -Suy --noconfirm base-devel git
+
+ARG AUR_PACKAGE=trunk-recorder-git
+
+RUN useradd builduser -m \
+  && passwd -d builduser \
+  && cd /home/builduser \
+  && git clone "https://aur.archlinux.org/$AUR_PACKAGE.git" target \
+  && chown builduser -R target \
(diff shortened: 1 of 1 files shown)


robotastic:
This awesome! I have a lot to learn about Arch linux.... I have no idea how AUR packages work. I wonder if doing a `make install` at the end would install the plugin libraries in the right place? I think the problem could be that the binary is getting moved, but not the plugin libraries. Doing an install might solve that.

toddejohnson:
Arch linux has one thing many other distros don't. The fact you can easily make a package of everything and share it in the Arch User Repositiory(AUR) and there are some wrappers to pacman(yay and others although I don't use them and they aren't supported by Arch Linux). 

Running linux I've hated installing unpackaged software and this makes getting it in a package simple. This also allows proprietary software to become packaged when installed. Rolling release is awesome and sucks(I have GR3.9 already but now I can't have GR3.8). Anyone can maintain an AUR package. I wasn't doing docker and outran my RPi so I made this one so I wasn't running unpackaged software on my system. 

I started with make install but struggled to get files in good places. You can see it still commented out. As AUR is user contributed you are encouraged to browse the [PKGBUILD](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=trunk-recorder-git) before building. Some of the wrappers include that as a step. I'm open to any needed changes in the PKGBUILDS and co-maintanership if you want.

PR #547, opened 2021-11-07
Readd support for reading in direct CSV exports from Radio Reference

devicenull:
The RadioReference.com talkgroup CSV format is not very different then the one trunk-recorder expects.  This patch adds support for using the CSV exported directly from RadioReference, while still supporting the custom trunk-recorder format.  It detects a RadioReference format file by looking for (and ignoring) the header line it inserts.


robotastic:
This is a  great idea! Could you include/attach a sample .csv file from radio reference, so I can test it out? I really have to upgrade sometime so I can download the .csv files.

devicenull:
Sure, sample attached.


[demo.csv](https://github.com/robotastic/trunk-recorder/files/7493919/demo.csv)

robotastic:
So - Slight wrinkle... in the talkgroups, the `mode` is a char variable. It can only work with a single character, so A, D, & E work, but TE is a problem. And I actually just checked my local system and they are using DE for encrypted channels. I will go work on changing it around so that `mode` is a string.

JoeGilkey:
Performance wise, making the mode an enum would be better.

robotastic:
Ok - I think I have things working. Merging in... thanks!
Can you check audioplayer.php? There was a conflict and I think I fixed things but may have messed it up.

PR #482, opened 2021-05-24
Per-day call log plugin, log P25 priority/mode/duplex

rosecitytransit:
Completes #153

https://github.com/robotastic/trunk-recorder/commit/9a115eba51594d5facf4edd3f1ec6fa66ff83c70 puts a link to my daily log plugin (and live Web page audio player) in README.md and makes a few minor changes, including correcting links

https://github.com/robotastic/trunk-recorder/commit/83bed1fc4a81ceb926b0300bdbb40522a49a6f9b adds logging of P25 call priority (can be 0-7), circuit/packet mode and duplex from the Service Options field in control channel packets. Was originally https://github.com/robotastic/trunk-recorder/commit/67a0d86428c3d625c3a607793c336c86169f69af

PR #549, opened 2021-11-09
Significant improvements to audioplayer

devicenull:
This is a fairly large change to audioplayer.php that adds the following features:

* Calls are retrieved via AJAX, and new calls are refreshed every 10s
* Audio playback will continue automatically when new calls are found
* Talk group selector was changed to use select2, which allows for searching the list of available call groups
* The file can now be installed anywhere - it does not require being in the "utils" directory (as long as you use absolute paths in config.json, and update audioplayer.php with the path to the config file)

I added a couple more Javascript libraries here... It took some effort to not rewrite this entire file to use a proper templating library and source control, but I figured keeping this as a single file was important

### DIFF
--- a/utils/audioplayer.php
+++ b/utils/audioplayer.php
@@ -1,11 +1,22 @@
 <?php
-$FileType = '.m4a';
+// no leading period
+$FileType = 'm4a';
 
-$system = (empty($_GET['system'])) ? null : $_GET['system'];
-$date   = (isset($_GET['date'])) ? new DateTimeImmutable($_GET['date']) : new DateTimeImmutable();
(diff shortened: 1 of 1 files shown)


rosecitytransit:
I actually have a "live" version that I made at https://github.com/rosecitytransit/trunk-recorder/tree/master/examples and discussed in #482

robotastic:
Thanks @devicenull - I will merge this in. 

@rosecitytransit - if you have some changes you want to make to the audio_player, that would be great. Could you open a PR for it? This is also a good reminder for me to look at the code in #482 and look into turning it into a plugin.

PR #551, opened 2021-11-10
Fix recording encrypted transmissions

tadscottsmith:
Based on the code, I believe you intended to ignore encrypted transmissions, but the bool is set backwards based on what p25_frame_assembler::make() expects.

### DIFF
--- a/trunk-recorder/recorders/p25_recorder_decode.cc
+++ b/trunk-recorder/recorders/p25_recorder_decode.cc
@@ -90,9 +90,9 @@ void p25_recorder_decode::initialize(  int silence_frames) {
   bool do_msgq = 0;
   bool do_audio_output = 1;
   bool do_tdma = 0;
-  bool do_crypt = 0;
+  bool do_nocrypt = 1;
 
-  op25_frame_assembler = gr::op25_repeater::p25_frame_assembler::make(0, silence_frames, wireshark_host, udp_port, verbosity, do_imbe, do_output, do_msgq, rx_queue, do_audio_output, do_tdma, do_crypt);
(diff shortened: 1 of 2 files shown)


robotastic:
Wow!! Nice catch, I think you are totally right. I am not sure how this snuck in there. I was looking through the OP25 commits and it look like it came in a while back:
https://github.com/boatbod/op25/commit/0eb22f06db54de76d7bee948360bfb49353221c1#diff-5cdfbb89a22cfc809bbbc39842c71ff33ef277237c8c9b8a5d0f6599899e057d

Here is the line where it is tested. If it is set to false, it will always record, even if it is encrypted.
https://github.com/robotastic/trunk-recorder/blob/b8ff5a5bd89e85e1e0a3642af13bb57ffd6a8c76/lib/op25_repeater/lib/p25p1_fdma.cc#L699

There weird thing is that a transmission shouldn't even start recording if the Grant message has the encrypted flag turned on. I wonder if it is being triggered by the GRANT and UPDATE trunking messages being treated the same... It could be worth trying to only start a recording on a GRANT message.

https://github.com/robotastic/trunk-recorder/blob/b8ff5a5bd89e85e1e0a3642af13bb57ffd6a8c76/trunk-recorder/main.cc#L953-L956

I will merge this in. Let me know if you are still having problems and I can try that GRANT/UPDATE thing

tadscottsmith:
There's definitely an issue with GRANT and UPDATE being handled the same way, as the UPDATE packet doesn't have the encryption information included. With some quick logging, it looks like on my system (Motorola P25 PI), I'm getting a number of recordings fired up off of UPDATE packets, when I would expect that to be near zero. On first glance, it looks like it may mostly occur when talk groups are patched, so I'll do some further testing to see what I can find.

PR #556, opened 2021-11-26
Add method in P25 Recorder to allow talkgroup ID to be retreived

aaknitt:
Allows audio_stream() in plugins to get talkgroup ID since Call info is protected and not available.


aaknitt:
Per Luke, this can be accomplished by having the plugin subscribe to the call_start() function.  The Call data will be available there, which includes both the TGID and the Recorder (and therefore the Recorder's unique ID that can be retrieved with get_num() ).  The plugin can then track which TGID is associated with a Recorder and act accordingly.  TGID and Recorder links can be removed in the call_end() function.

PR #557, opened 2021-11-26
Plugin rdioscanner

chuot:
Here is the Rdio Scanner plugin that I have been testing since last week without any problem. It has been tested with Rdio Scanner v6.0, but should work fine with v5.2.x, which is the current stable version.

With this plugin, Rdio Scanner (v6.0) will automatically populate it's systems configuration based on what's in the talkgroup file, if any.

Hopefully this will make things easier for Rdio Scanner users.

If I'm missing something, don't hesitate!

Thanks for your wonderful project !!!


robotastic:
This is awesome! Ok - let go through the code. I think it might be easier to move the config settings for it to the plugins section. It had kept broadcastify as part of the main config because it had already been in there and I wanted to minimize the amount people had to change.

Let me see if I can go make the changes. This is the first plugin and it is so cool to see people building on this. Rdio Scanner is great and this should make setup a lot easier.

robotastic:
I moved the config section for the plugin over into the plugin section of the config.json file. This should let you keep all the changes you need to make to your folder and make it easier to do PR in the future.

I still need to update to documentation for this and give it one more look over. 

Here is an example of what you would add the config.json to enable the plugin:

```
"plugins": [
        {
          "name": "rdioscanner_uploader",
          "library": "librdioscanner_uploader.so",
          "server": "http://127.0.0.1",
          "systems": [{
                  "shortName": "test",
                  "apiKey": "fakekey",
                  "systemId": "411"
          }]
        }
],
```

robotastic:
I moved the config section for the plugin over into the plugin section of the config.json file. This should let you keep all the changes you need to make to your folder and make it easier to do PR in the future.

I still need to update to documentation for this and give it one more look over. 

Here is an example of what you would add the config.json to enable the plugin:

```
"plugins": [
        {
          "name": "rdioscanner_uploader",
          "library": "librdioscanner_uploader.so",
          "server": "http://127.0.0.1",
          "systems": [{
                  "shortName": "test",
                  "apiKey": "fakekey",
                  "systemId": "411"
          }]
        }
],
```

robotastic:
I update the documentation and changed the config around. I am going to merge this in. Can you give it a test when you have a chance? I don't have Rdio Scanner setup yet.

chuot:
Thank you so much @robotastic!

Rdio Scanner v6 will be released this week in a beta branch, but the plugin should work with the latest v5 with some limitations in regards to the auto populate option.

taclane:
I know that Rdio Scanner v6 is on the horizon, but it doesn't seem like this plugin works well too with 5.2.9.  

The plugin uploads fine, but there's some underlying dateTime compatibility issue in 5.2.9 requiring me to manually prune the test entries from the sqlite database or it effectively breaks the app by sorting all the undated calls ahead of any dated ones.
Screen Shot 2021-12-16 at 5 37 48 PM

PR #559, opened 2021-11-27
Plugin Audio Streaming CPU/Memory Usage Reduction

JoeGilkey:
This change allows the audio streaming callbacks to be turned off if they are not being used. This reduces the memory and CPU usage. On RaspberryPi3 and other small devices, this can add significant overhead on a feature that is not commonly used.


robotastic:
Thanks so much for contributing the Plugin code - it has been working out great. Hopefully I didn't mess things up too much when I integrated it.

This fix is a great idea. Should it be a single setting at the top level for the whole instance? Or done system by system?

JoeGilkey:
No problem. I switched jobs and haven't had time to work on anything else for a while. I've got some time now, so I figured I would take a look. It looks great, and I have a few ideas that may help (i.e. when to delete the wav files).

For this I think you are correct and it should be a global flag. Individually, it should be on the system, not the source. Global makes more sense.

robotastic:
@aaknitt  heads up on this incoming change. Streaming audio will be off by default.

PR #560, opened 2021-11-29
Fix compile errors on Fedora 35

chuot:
Compiling on Fedora 35 broke since e83db81, here's a fix.


chuot:
So it hasn't passed the buildx process. I guess that the gnuradio from the buildx process isn't the same version as gnuradio-devel-3.9.2.0-7.fc35.x86_64 which is the latest version on F35.

robotastic:
Ah! I think I actually have to add in an IFDEF to test the gnuradio version and correct things. I must have missed these changes when I was importing a newer version of OP25. Let me go check.

PR #568, opened 2021-12-03
Update nonstop_wavfile_sink_impl.cc

tadscottsmith:
Change wavfile sink logLevel to trace.


tadscottsmith:
This is currently writing a ton of non-error messages when the logLevel is set to error.

robotastic:
Good call - that is more of a debug message.

PR #570, opened 2021-12-07
Conventional P25 demod checks after analog transmissions

taclane:
As discussed in #318 and #322, a Conventional P25 recorder often locks up when an analog transmission such as a morse ID occurs on that frequency.

Currently, the costas clock is reset at the beginning of each call.  But when analog transmissions occur *anywhere afterward* to knock the costas clock off, no further audio is decoded, so `get_current_length()` in `main.cc:void manage_conventional_call()` will not be > 0, and the call will not time out or conclude.  The recorder remains stuck until it either catches a lucky fragment of garbled P25 audio (`get_current_length() > 0`), or until trunk-recorder is restarted.

My local systems aren't very active, but I used to restart trunk-recorder via a cron job twice an hour because of CW IDs.  After this, it's been handling station IDs and other spurious analog transmissions over the past couple weeks and resetting the costas timer as required.

There may be other ways to address this, but looking for instances where the squelch opens and closes without producing audio seems to be a simple indicator that the Conventional P25 demod should probably be reset, and doesn't interfere with trunked systems I can't test against.

### DIFF
--- a/trunk-recorder/main.cc
+++ b/trunk-recorder/main.cc
@@ -725,11 +725,9 @@ void process_message_queues() {
   }
 }
 
-void manage_conventional_call(Call *call) {
-  if (call->get_recorder()) {
-    // if any recording has happened
 
-    if (call->get_current_length() > 0) {
+
+void handle_conventional_call_recording(Call *call) {
       BOOST_LOG_TRIVIAL(trace) << "[" << call->get_short_name() << "]\t\033[0;34m" << call->get_call_num() << "C\033[0m Call Length: " << call->get_current_length() << "s\t Idle: " << call->get_recorder()->is_idle() << "\t Idle Count: " << call->get_idle_count();
 
       // means that the squelch is on and it has stopped recording
@@ -762,6 +760,16 @@ void manage_conventional_call(Call *call) {
           plugman_setup_recorder(recorder);
         }
       }
+}
+
+void manage_conventional_call(Call *call) {
+
+  // make sure there is an associated recorder... there better be, since it is conventional
+  if (call->get_recorder()) {
+
+    // if any recording has happened
+    if (call->get_current_length() > 0) {
+      handle_conventional_call_recording(call);
     } else if (!call->get_recorder()->is_active()) {
       // P25 Conventional Recorders need a have the graph unlocked before they can start recording.
       Recorder *recorder = call->get_recorder();
--- a/trunk-recorder/main.cc
+++ b/trunk-recorder/main.cc
@@ -739,6 +739,9 @@ void manage_conventional_call(Call *call) {
       } else if (call->get_idle_count() > 0) {
         // if it starts recording again, then reset the idle count
         call->reset_idle_count();
+      } else if (call->get_recorder()->get_costas_error_count() > 0) {
+        // conventionalP25 recieving audio; no need to reset the demodulator if indicated
+        call->get_recorder()->reset_costas_error_count();
       }
 
       // if no additional recording has happened in the past X periods, stop and open new file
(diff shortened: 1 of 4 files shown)


robotastic:
Ah - so the part I was missing is that calling costa_reset on call_start() doesn't help much because with conventional, it will still be sitting around for a while until the squelch unlocks.

I like this approach - it seems pretty practical. My only concern is if too much stuff is getting jammed into the manage_conventional_calls() function. There are a lot of IFs. I might see if I can break it up a little.

taclane:
Heh heh. Sorry if the IFs are a little clunky.  Honestly, if the only change to manage_conventional_calls() is adding something like this on the end:
```
} else if (!call->get_recorder()->is_squelched()) {
  call->get_recorder()->costas_reset();
}
```
it still works, even if it is simply brute-forcing resets when the squelch opens up and isn't generating audio.  All the other lines to track how long it unsquelched was just nice because it made it easy to spot what was going on with each receiver.
```
(info)   [test]	1C	Conventional P25 squelch open 10s without decoded audio. Demodulator reset.
```
By me, 10s bursts are station IDs, 6s are DTMF tones, etc...

Resetting the demod at the start of each call still seems helpful.  But the problem was mostly that if the costas degrades sometime later, manage_conventional_calls() just wound up permanently stuck.  Regardless of anything that followed, get_current_length() could not go above 0 until the demodulator was reset again.  And the call couldn't restart and reset the modulator as long as get_current_length() was still 0.

robotastic:
Ok! I did a little refactoring... but I also may have found an alternate solution. When a Squelch block unlocks, it sounds out a message:

> The block will emit a tag with the key pmt::intern("squelch_sob") with the value of pmt::PMT_NIL on the first item it passes, and with the key pmt::intern("squelch:eob") on the last item it passes.

https://www.gnuradio.org/doc/doxygen/classgr_1_1analog_1_1pwr__squelch__cc.html

I changed the Costas block to look for that and automatically reset each time it see that message. I don't have any way to test if this works. Could you give it a try? I added in a debug message to print out when this happens. The branch is:
`auto-reset` and the code for this is here:
https://github.com/robotastic/trunk-recorder/blob/39cafa8e5e2d12515c1c6f168910029fb74997d8/lib/op25_repeater/lib/gardner_costas_cc_impl.cc#L270-L285

taclane:
Great! The `auto-reset` does does seem promising, though I've already seen a few instances like:
```
[2021-12-08 00:05:05.005838] (info)   [  ]
[2021-12-08 00:05:05.005864] (info)   [ A2 ] State: idle
[2021-12-08 00:06:04.129350] (trace)   P25 Frame Assembler - Amt Prod: 864 output_queue: 0 noutput_items: 864
[2021-12-08 00:06:04.233832] (trace)   P25 Frame Assembler - Amt Prod: 864 output_queue: 0 noutput_items: 864
[2021-12-08 00:06:04.337098] (trace)   P25 Frame Assembler - Amt Prod: 864 output_queue: 0 noutput_items: 864
```
where it looks like it missed the unsquelch and didn't reset before a transmission.  It catches things most of the time, but I'm going to have to get a second trunk-recorder going to figure out if there's any pattern to why it isn't resetting sometimes.  I'll let them both run for a while and see how it goes.

robotastic:
Oh interesting.... it could be that resetting it is causing it to miss some normal transmission that it would have normally decoded. If you are able to find a pattern, that would be really helpful. It may end being better to use your technique, which only resets the Costas block when it stops decoding correctly.

taclane:
Here are some observations after letting them both run over 24 hours on identical equipment, feeding the same instance of rdio-scanner:

**Auto-reset fix**
- On rare occasions, it doesn't log a reset on unsquelch.  Not sure what the root cause is, but it only seems to happen during analog transmissions.  Subsequent P25 transmissions are still resetting on unsquelch.  It doesn't appear any P25 audio was ever missed.
- Resetting on open squelch catches analog transmissions that happen in the middle of a call.  If an analog transmission happens in the middle of a P25 call, it will still record everything afterward.  It can handle a call that looks like:
```(P25 Audio) -> (Analog Station ID) -> (P25 Audio) -> (P25 Audio)```
when the other method would only produce
```(P25 Audio) -> (Station ID) -> (Unable to decode P25 until the call times out)```

**manage_conventional_call() fix**
- It is really helpful for troubleshooting to see which recorder is having issues with undecoded audio.  This:
```[test]	1C	Conventional P25 squelch open 9s without decoded audio. Demodulator reset.```
is much more useful to diagnose than hundreds of:
```P25 Frame Assembler - Amt Prod: 864 output_queue: 0 noutput_items: 864```
```P25 Frame Assembler - Amt Prod: 864 output_queue: 0 noutput_items: 864```
```P25 Frame Assembler - Amt Prod: 864 output_queue: 0 noutput_items: 864```

I'm leaning toward `auto-reset` being a better solution.  It seems like the simpler fix, and addresses some things that the other method can't.  Both are effective workarounds for analog transmissions, and each successfully recorded the same number of conventional P25 calls without requiring manual intervention.

**Comment by taclane** [comment]
I added a second check in `auto-reset`, and it seems like the few times it didn't detect a squelch opening, it managed to find a `squelch_eob` tag at the beginning of those transmissions.
```
[2021-12-09 13:22:59.005691] (info)   [ A2 ] State: idle
Detected a Squelch Close, resetting the Costas Clock!
[2021-12-09 13:24:28.634448] (trace)   P25 Frame Assembler - Amt Prod: 864 output_queue: 0 noutput_items: 864
```
I'm not sure if it could cause problems for other systems, but resetting the costas timer for *any* of the squelch tags seems to guarantee that a transmission always starts off with a ready demodulator:
```
  std::vector<gr::tag_t> tags;
  std::vector<gr::tag_t> tags_squelch_eob;
  pmt::pmt_t squelch_key(pmt::intern("squelch_sob"));
  pmt::pmt_t squelch_end_key(pmt::intern("squelch_eob"));
...
  get_tags_in_range(tags, 0, nitems_read(0), nitems_read(0) + noutput_items, squelch_key);
  get_tags_in_range(tags_squelch_eob, 0, nitems_read(0), nitems_read(0) + noutput_items, squelch_end_key);
  if (!tags_squelch_eob.empty()||!tags.empty() ) {
      fprintf(stderr, "Detected a Squelch event, resetting the Costas Clock!\n");
      reset();
  }
  tags.clear();
  tags_squelch_eob.clear();
```

robotastic:
Cool! Good idea on check for the end of squelch too. It is weird that the message doesn't get sent out sometime. I wonder if there are some instances when the Squelch block hasn't fully closed. I am not quite sure how attack and decay come into play. I looked through the code and didn't see anything obvious:
https://github.com/gnuradio/gnuradio/blob/master/gr-analog/lib/squelch_base_ff_impl.cc

Let's go with the `auto-reset` approach for now. It is a bit more straightforward and less likely to break if I make changes somewhere down the road.

In theory we could add a debug message in P25 Frame Assembler that prints out when an 864 byte sample is received but nothing is decoded.

I will go make those changes to the brnach.

taclane:
Closing the PR since it looks like there's a path forward on this with the `auto-reset` branch.  I'll post any new issues or discoveries back in #318.

PR #571, opened 2021-12-09
simplestream Plugin

aaknitt:
Plugin to stream raw uncompressed audio from all or specified talkgroups being recorded via one or more UDP ports.  Allows other downstream programs to play audio in real time or to compress and send to streaming services.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -297,6 +297,8 @@ add_subdirectory(plugins/unit_script)
 
 add_subdirectory(plugins/rdioscanner_uploader)
 
+add_subdirectory(plugins/simplestream)
+
 add_executable(trunk-recorder trunk-recorder/main.cc ${trunk_recorder_sources})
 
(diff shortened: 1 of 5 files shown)


aaknitt:
This is going to fail to build until the patch following stuff gets merged...I designed the plugin to use that data which doesn't yet exist in the master branch.

robotastic:
This is great!! It is based on having the Patch code merged in first, right?

A thought - since this is really 16-bit audio, I could move the float to int16 conversion out of the wav_file block and earlier in the process, before the plugin. I might be wrong, but this might cut the bandwidth req in half and the audio should sound the same.

aaknitt:
> This is great!! It is based on having the Patch code merged in first, right?

Yes, I got ahead of myself a bit and put patch following into this plugin so that it'll stream patched talkgroups.  So it's not going to work until the patch following code is in first.

> A thought - since this is really 16-bit audio, I could move the float to int16 conversion out of the wav_file block and earlier in the process, before the plugin. I might be wrong, but this might cut the bandwidth req in half and the audio should sound the same.

Yeah that would cut the bandwidth in half, I like it!

robotastic:
Ok - I think I made the switch... checkout the `short-wav` branch. 

Actually, that could be tricky, since it doesn't have the Patch Fork merged in yet.

aaknitt:
> Ok - I think I made the switch... checkout the `short-wav` branch.
> 
> Actually, that could be tricky, since it doesn't have the Patch Fork merged in yet.

Seems to be working!  One gotcha is that I had to dial back my digitalLevels setting in config.json after making this change.  I think I can understand why that may be necessary, and I don't think it hurts anything other than requiring config file changes when this change hits?

If it makes things easier, I can merge these changes into the PR for the patch following stuff and close this PR.

robotastic:
We can just keep it as a distinct PR. If you think things are good, I can merge the different PRs and branches together this weekend.

PR #575, opened 2021-12-13
Update main.cc - fix for segfault on conventional systems

aaknitt:
In a conventional system the sys object isn't initialized since a control channel message is never received, leading to a segfault when the clear_stale_talkgroup_patches() function gets called and gets passed an empty System object.  This check prevents calling the function unless the system type is known as p25, which means by definition that it's not empty.


robotastic:
I fixed the root cause ... should be good now.

aaknitt:
Got it, makes sense.  I forget that there can be more than one system since I only ever use one.  Sorry for the oversight!

robotastic:
No Probs!! Thanks for all the work pulling it together!!

PR #579, opened 2021-12-14
Talkgroup patching priority fix

aaknitt:
Prevent priority increases from talkgroup patching from being permanently stored in the System->Talkgroup objects


robotastic:
This makes sense... eventually we may need to do something similar for analog talkgroups if Smartnet ever gets figured out, but lets figure that out later.

PR #587, opened 2021-12-21
Update talk group patches to use unsigned long throughout the app.

tadscottsmith:
P25_parser.cc was writing talk goups using unsigned long data types, and most of the patch tracking was using signed long data types. This could lead to unexpected results.


robotastic:
Good catch!!

PR #588, opened 2021-12-24
simplestream plugin segfault fix, update_active_talkgroup_patches() change

aaknitt:
Added mutex to fix segfaults in simplestream plugin.  Been running well for ~48 hours vs. segfault after several hours.

Reduced amount/type of data being passed into update_active_talkgroup_patches().


aaknitt:
Still having issues, closing until I get it fully resolved.

Issue #590, opened 2021-12-26
Segfault in Talkgroup::get_priority


patrickfish:
Not sure if related to #585 since I'm not using simplestream, but I've been getting this every so often and finally caught it.

```
[Thread 0x7ffdba604700 (LWP 21387) exited]
[2021-12-25 17:40:48.009935] (error)   [system]       49784C Failed to conclude call - TG:       3847
        Sat Dec 25 17:37:26 2021 PST
[New Thread 0x7ffdba604700 (LWP 21446)]
[New Thread 0x7ffdb8600700 (LWP 21447)]
[Thread 0x7ffdb8600700 (LWP 21447) exited]

Thread 1 "trunk-recorder" received signal SIGSEGV, Segmentation fault.
0x000055555569afd2 in Talkgroup::get_priority (this=0x0)
    at /home/pfish/trunk-recorder4/trunk-recorder/talkgroup.cc:29
29        return priority;
(gdb)
(gdb)
(gdb)
(gdb)
(gdb) bt full
#0  0x000055555569afd2 in Talkgroup::get_priority (this=0x0) at /home/pfish/trunk-recorder4/trunk-recorder/talkgroup.cc:29
No locals.
#1  0x00005555555e0f66 in start_recorder (call=0x55555b251d10, message=..., sys=0x555559f99a80) at /home/pfish/trunk-recorder4/trunk-recorder/main.cc:642
        TGID = @0x7fff18004ec0: 3009
        _foreach_continue641 = false
        _foreach_end641 = @0x7fffffffd098: {<No data fields>}
        _foreach_cur641 = @0x7fffffffd090: {<No data fields>}
        _foreach_col641 = @0x7fffffffd120: {<No data fields>}
        priority = 1
        source = 0x555559fb3270
        it = 0x555559fb3270
        talkgroup = 0x555559faa400
        source_found = true
        recorder_found = false
        recorder = 0x2d
        debug_recorder = 0x55555b251d10
        sigmf_recorder = 0x55555563629a <boost::algorithm::detail::is_classifiedF::~is_classifiedF()+28>
#2  0x00005555555e3b68 in handle_call (message=..., sys=0x555559f99a80) at /home/pfish/trunk-recorder4/trunk-recorder/main.cc:962
        call = 0x55555b251d10
        call_found = false
        call_retune = false
        recording_started = false
[...]

PR #591, opened 2021-12-27
Update main.cc - add NULL check for segfault issue

aaknitt:
https://github.com/robotastic/trunk-recorder/issues/590


robotastic:
Good catch - This should catch when a TG that is part of a Patch is not part of the talkgroup file

PR #597, opened 2022-01-06
Add patch info to rdio-scanner API in plugin

aaknitt:


robotastic:
@chuot Since you contribute this plugin, could you take a look at the proposed change?

chuot:
@robotastic I've been running it for a full day without any issue. It can be merged. Thanks!

PR #600, opened 2022-01-06
Add VGA and AMP gain stages for Soapy-Remote users

mepholic:
This adds 2 new gain stages that you can set from the configuration file `VGA` and `AMP`. These are seemingly needed to set the Gain properly on my HackRF that I'm interacting with using Soapy-Remote.

The source is configured with the following properties to get that to work:
```
"lnaGain": 16,
"vgaGain": 48,
"ampGain": 14,
"driver": "osmosdr",
"device": "soapy=0,driver=remote,remote=192.168.1.20,remote.driver=hackrf"
```

### DIFF
--- a/trunk-recorder/main.cc
+++ b/trunk-recorder/main.cc
@@ -428,6 +428,8 @@ bool load_config(string config_file) {
       int lna_gain = node.second.get<double>("lnaGain", 0);
       int pga_gain = node.second.get<double>("pgaGain", 0);
       int tia_gain = node.second.get<double>("tiaGain", 0);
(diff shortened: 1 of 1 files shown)


robotastic:
Cool - that works. There was recently an addition that makes it easy to add gain stages with arbitrary names, but it can't hurt to also have first order support for them:

gainSettings  | set the gain for any stage. The value for this setting should be passed as an object, where the key specifies the name of the gain stage and the value is the amount of gain, as an int. For example:"gainSettings": { "IF": 10, "BB": 11},

PR #572, opened 2021-12-10
Correct the squelch key for squelch_eob

taclane:
Since I was investigating other squelch-related stuff for #318 and #570, this looks to be a typo and will not be found by the nonstop_wavfile_sink.

While the documentation for pwr_squelch does say:
> The block will emit a tag with the key pmt::intern("squelch_sob") with the value of pmt::PMT_NIL on the first item it passes, and with the key pmt::intern("squelch:eob") on the last item it passes.

It appears to be in error, as the underlying squelch_base passes `squelch_eob` upon closing, not `squelch:eob`.

https://github.com/gnuradio/gnuradio/blob/a2a682518fab2a91b5b4628586568f2e1cf1b65b/gr-analog/lib/squelch_base_ff_impl.cc#L27-L28

PR #608, opened 2022-01-16
Revert 572 squelch eob typo

taclane:
Reverts robotastic/trunk-recorder#572

Ultimately, things were better when squelch_eob _wasn't_ working.

Terminating a transmission based on squelch_eob leads to dropped audio samples for conventional P25 (and presumably conventional DMR) when a unit's transmission ends up spread across multiple .wav files.  I'm not sure if this change even had an effect on trunked systems, as they don't rely on squelch as heavily as conventional digital modes.


robotastic:
Sounds good - it was worth a shot!

PR #618, opened 2022-01-21
Update rdioscanner_uploader.cc

taclane:
As discussed in https://github.com/chuot/rdio-scanner/issues/78, the rdio-scanner plugin does not currently check if it is using compressed audio when it reports a call filename to the server.  Audio sent via the plugin is logged, stored, and downloaded with a .wav extension, even when the actual content inside may be m4a compressed.

Uploading the with the correct filename should help with those issues, and potentially resolve other problems such as file corruption when rdio-scanner attempts to compress the incoming audio a second time due to the .wav extension.


robotastic:
Great catch!
@chuot - Tagging you here for awareness

Issue #613, opened 2022-01-18
Add system demarcation to simplestream plugin


aaknitt:
Via #simplestream Discord channel:

simplestream plugin needs system demarcation  to prevent TG n from one system colliding with the same numbered TG from another system.

Proposed solution is to add a system key to the streams object (separation would need to be handled by the user via UDP port separation), so configs might look like:
```
{
  "name": "simplestream",
  "library": "libsimplestream.so",
  "streams": [
    {"system": "system1", "TGID": 1234, "address": "127.0.0.1", "port": 1234, "sendTGID": false},
    {"system": "system2", "TGID": 1234, "address": "127.0.0.1", "port": 1235, "sendTGID": false},
  ]
}
```
I'll work on this, just adding it here for tracking & feedback/comments.

aaknitt:
Testing here:
https://github.com/aaknitt/trunk-recorder/tree/simplestream_add_system

PR #620, opened 2022-01-25
Simplestream updates to support conventional recorders, add System demarcation

aaknitt:
https://github.com/robotastic/trunk-recorder/issues/613

Allows differentiation between systems that share common TGIDs.  

Eliminates removing TGID mapping in conventional recorders when call_end() is called since same recorder is used for duration of program execution.


robotastic:
Good catch on the sample rates being different for Analog and Digital!

Issue #615, opened 2022-01-18
audio_type is wrong in json


ZeroChaos-:
json
```
{
"freq": 460050000,
"start_time": 1642520058,
"stop_time": 1642520062,
"emergency": 0,
"encrypted": 0,
"call_length": 5,
"talkgroup": 1,
"talkgroup_tag": "Moon Police",
"audio_type": "digital",
"short_name": "moonpd",
"freqList": [{ "freq": 460050000, "time": 1642520058, "pos": 0.0, "len": 5, "error_count": 0.0, "spike_count": 0.0}],
"srcList": [ {"src": -1, "time": 1642520058, "pos": 0, "emergency": 0, "signal_system": "", "tag": ""} ]
}
```
config snippit showing the system
```
      {
        "channels": [460050000],
        "alphatags": ["Moon Police"],
        "talkgroupDisplayFormat": "tag_id",
        "type": "conventional",
        "callLog": true,
        "audioArchive": true,
        "minDuration": 1,
        "shortName": "moonpd",
        "talkgroupsFile": "moonpd.csv",
        "analogLevels": 12,
        "decodeMDC": "true",
        "uploadScript": "encode-upload.sh",
        "unitTagsFile": "units.csv",
        "compressWav": false,
        "squelch": -23
      },
```

ZeroChaos-:
Unfortunately, still wrong for conventional analog channels
```
{
"freq": 460050000,
"start_time": 1642782358,
"stop_time": 1642782385,
"emergency": 0,
"encrypted": 0,
"call_length": 28,
"talkgroup": 1,
"talkgroup_tag": "Moon Police",
"audio_type": "digital",
"short_name": "moonpd",
"freqList": [{ "freq": 460050000, "time": 1642782358, "pos": 0.0, "len": 28, "error_count": 0.0, "spike_count": 0.0}],
"srcList": [ {"src": 86, "time": 1642782358, "pos": 0, "emergency": 0, "signal_system": "", "tag": "Moon PD Portable 56"} ]
}
```

PR #621, opened 2022-01-26
Set set_is_analog(true) for conventional analog

taclane:
Per https://github.com/robotastic/trunk-recorder/issues/615, conventional analog shows up as digital in json logs since `set_is_analog` never set and the call reports back the default value of false.
```
"audio_type": "digital",
```

Adding `call->set_is_analog(true)` to conventional system setup correctly identifies those calls as analog in the resulting json.
```
"audio_type": "analog",
```


robotastic:
Yep! I think that should fix it. Thanks for tracking this down!

PR #624, opened 2022-01-29
Add info to Recorder object for stream_audio()

aaknitt:
- Add missing function calls to populate talkgroup in all recorder types (already existed in analog but missing in P25 and DMR)
- Add System short_name to info in Recorder object

Both of these items can be useful in the stream_audio() function in plugins to provide relevant metadata that matches the audio samples being passed in.

### DIFF
--- a/trunk-recorder/recorders/analog_recorder.cc
+++ b/trunk-recorder/recorders/analog_recorder.cc
@@ -84,6 +84,7 @@ analog_recorder::analog_recorder(Source *src)
   talkgroup = 0;
   recording_count = 0;
   recording_duration = 0;
+  short_name = "";
 
   rec_num = rec_counter++;
   state = INACTIVE;
@@ -334,6 +335,11 @@ long analog_recorder::get_talkgroup() {
   return talkgroup;
 }
 
+std::string analog_recorder::get_short_name(){
+  return short_name;
+}
+
+
 double analog_recorder::get_freq() {
   return chan_freq;
 }
@@ -398,6 +404,7 @@ bool analog_recorder::start(Call *call) {
 
   talkgroup = call->get_talkgroup();
   chan_freq = call->get_freq();
+  short_name = call->get_short_name();
 
   squelch_db = system->get_squelch_db();
   squelch->set_threshold(squelch_db);
--- a/trunk-recorder/recorders/analog_recorder.h
+++ b/trunk-recorder/recorders/analog_recorder.h
@@ -80,6 +80,7 @@ public:
   void set_source(long src);
   Source *get_source();
   long get_talkgroup();
+  std::string get_short_name();
   time_t get_start_time();
   double get_current_length();
   long get_wav_hz(); 
@@ -104,7 +105,7 @@ private:
   double center_freq, chan_freq;
   long talkgroup;
   long samp_rate;
-
+  std::string short_name;
   double system_channel_rate;
   double initial_rate;
   float quad_gain;
--- a/trunk-recorder/recorders/dmr_recorder.cc
+++ b/trunk-recorder/recorders/dmr_recorder.cc
@@ -160,6 +160,7 @@ void dmr_recorder::initialize(Source *src) {
   squelch_db = 0;
 
   talkgroup = 0;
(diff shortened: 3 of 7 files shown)


robotastic:
I am not sure if this is the correct approach. In setups where there is more than one system, it is possible for a recorder to be used across multiple systems. It is better to check the call information to determine what system a recorder is currently associated with.

aaknitt:
@robotastic I'm not sure I'm tracking with you.  I agree and understand that a single Recorder can be used across multiple Systems, which is why I'm suggesting adding the System shortName to the data held in the Recorder.  By having this info contained in the Recorder, the logic in a plugin for streaming audio can be substantially simplified.  Rather than having to use the call_start() and call_end() functions to track (within the plugin) which Recorder is currently associated with a particular System and/or TGID, all that's needed is some simple logic checks within the audio_stream() function using the Recorder info that's passed in: Is this Recorder associated with a System+Talkgroup I care about?  If so, do something; if not, ignore it.  

My thought is that since trunk-recorder is already tracking which System+Talkgroup a Recorder is associated with, it may make sense to pass this info to the plugin in the Recorder object so that trunk-recorder and the plugin aren't both trying to track the same information in parallel (and independently, which is prone to errors in the plugin).

If I'm thinking about this correctly (I very well may not be) this could enable a greatly simplified rewrite of simplestream and make future streaming plugins easier to develop.  

On the other hand, simplestream seems to be working well as-is, so if you'd prefer to continue with the current approach it's no big deal, I was just thinking this could make life easier for others wanting to write plugins with similar functionality. Does adding this info potentially break something or is there opportunity for it to be incorrect?  

By way of example, here's a very simple audio_stream() function for a plugin that simply prints out the info that I've added to the Recorder in this PR:
```
  int audio_stream(Recorder *recorder, int16_t *samples, int sampleCount){
    BOOST_LOG_TRIVIAL(info)<<"recorder "<<recorder->get_num()<< " is on freq "<<recorder->get_freq()<< " for TGID "<<recorder->get_talkgroup()<<" on system "<<recorder->get_short_name();
	return 0;
  }
```
Here's a log statement showing Recorder 3 being used on one system:
```
[2022-01-29 20:25:41.294335] (info)   recorder 3 is on freq 1.511e+08 for TGID 57511 on system washingtonvhf
```
and then later in the same log file showing the same Recorder 3 being used on a different system:
```
[2022-01-29 20:26:32.588779] (info)   recorder 3 is on freq 1.5284e+08 for TGID 58903 on system WISCOMdelafield
```
So it seems like the info is making it into the Recorder correctly, as best I can tell.

robotastic:
Sorry - I hadn't actually be tracking the streaming plugin recently. I didn't realize that Stream hook only passed the recorder. I am a little nervous to start tracking state in a couple different places. I know it seems a little  silly, but it helps because I end up forgetting to about update things as the code changes. What we just passed the call along with the recorder? It is easy to pull the short_name and talkgroup from that?

aaknitt:
Yeah passing in the Call along with the Recorder would definitely make everything that's desired (and then some) easily available.  If that's easy to do, I agree that's a better solution.

aaknitt:
Does this look like what you're thinking?

Add something like this to the p25_recorder_decode.h, analog_recorder.h, and dmr_recorder.h:
image
And then something like this to the associated start() functions:
image
And then it can get passed to the plugin here:
image

robotastic:
Cool! I think it might take a little more work, but it shouldn't be bad. I can go make that change tonight.

robotastic:
ok - I gave it a go and created the `plugin-stream` branch... it compiles, but I  haven't had time to run it. It should be passing the Call object to the audio streamer now.
0167d947a98077d4b06e4667200479f25eab38c8

aaknitt:
Awesome, seems to be working well.  Looks like the DMR recorder isn't calling the audio_stream() function for some reason, not sure if that's on purpose or not.  

I'm going to go ahead and close this PR.  Once the plugin-stream branch is merged to main I'll probably tackle a simplified rewrite of simplestream but no rush on that as it won't be a functional change, just cleanup to take advantage of this.

Issue #100, opened 2017-03-02
Add Support for Digital Coded Squelch


kkotowick:
I'm using this software to record some conventional channels. The radio network uses digital coded squelch (DCS) so that receivers don't hear a bunch of non-verbal traffic that is transmitted over the channel (morse code for the station ID, etc.). Unfortunately, using your software, it captures all this non-verbal transmission as well.

What are the chances of getting a feature added that allows us to put a receive squelch code in the config file (usually 3-digit octal), so that it only records audio when the DCS is present?

Thanks!

robotastic:
I looked into this a little while ago and I think it might be possible. I found a project that already implemented most of the code:
https://bitbucket.org/holiman/gnuradio-dcs
I will add this on to the list of features to add.

kkotowick:
Awesome, thanks! One thing to keep in mind: it will have to be able to detect the presence of the DCS even if it appears in the middle of an ongoing transmission (i.e. can't just check once when a new transmission starts, then ignore the rest of the transmission if it doesn't have it).

Dygear:
This along with tone squelch is the would make this an almost competle system for me. Recording the DCS / TSQ values would be very handy. I actually have a frequency that is used by two different departments that are active in the same area, but the 2nd department uses a different tone. So it would be cool if we could setup the system like this ...

```
"systems": [{
  "type":     "conventional",
  "channels": [151070000: [{"DCS": 071}], 154115000: [{"TSQ": 100.0], 154570000: [{"TSQ": 151.4}]]
}, {
  "type":     "conventional",
  "channels": [460175000: [{"DCS": [315, 243]}], 461175000: [{"DCS": 516}], 467725000: [{"DCS": 043}]]
}]
```

And the tones that were detected and present within the waveform, it would be perfect if this information also exposed to the outlets (JSON, API, WebSocket).

kkotowick:
@robotastic You had mentioned you had looked into this a while back. Any luck with implementing it? It would be an amazing enhancement for conventional systems, so it wouldn't capture all of the unwanted traffic.

imagesafari:
+1 for tone squelch

imagesafari:
Bump for DCS / CTCSS squelch! :)

Cole0209:
Agreed this would be great

imagesafari:
Bump

EricTendian:
[...]

PR #630, opened 2022-02-04
Conventional talkgroup number and alpha text assignments using .csv files

taclane:
Conventional channels are currently assigned sequential talkgroup numbers based on their order in the config. Because of the importance of those talkgroup numbers downstream, it is difficult to rearrange, add, or subtract conventional frequencies in a system without potentially breaking things or causing mixups with alpha tags

This PR proposes one solution through the use of the `talkgroupFile` csv as a lookup table for conventional channels by:
- Storing the channel frequency in the unused second field of the talkgroup csv
- Overriding any json Alpha Tags if a conventional system has a talkgroupFile set
- Searching for a matching talkgroup by channel frequency and using the assigned TG ID and alpha tags in system setup

If a talkgroup file is not listed for a conventional system, it falls back on sequential TG numbers and alpha tags.

Aside from the talkgroup class including a new field, there has been no additional impact noted on trunked systems. The block of code managing this only exists within the conventional system setup.

`config-conv-tg-csv.json`, `conv-tg.csv`, and `frs.csv` are provided in /examples to show how this might work.


robotastic:
This looks pretty good! How about we create a new file format just for conventional channels though? I would rather do that then have fields in Talkgroup which are overloaded. We could also add in the Tone information in the future. #100 
Are there fields in the current Talkgroup file format that we could ditch for Conventional?

taclane:
Ideally, a conventional system can benefit from almost all the same fields that a trunked system can, such as:
`TG#, Alpha Text, Description, Tag, Group`
Especially with something Rdio Scanner downstream, getting all those text fields to auto populate makes things easy to manage and display.

I agree that a tone field should be reserved for future use. Being able to copy something like like "026 DPL" or "107.2 PL" or "CSQ" from Radio Reference could make it pretty simple.  Having frequency or frequency/tone to look up call information should cover most cases.  

There's been a lot of discussion about conventional channels on here and the discord lately, so it's probably worth gathering some additional input to see what other scenarios may need to be covered.

robotastic:
Just to keep things updated - I am working on branch that will have most of this functionality. Check out the `conventional` branch. It is almost ready for testing.

robotastic:
OK! This should be added with #652 - I am going to close this PR. Thanks for the great idea!

PR #631, opened 2022-02-05
simplestream plugin update

aaknitt:
major simplification taking advantage of Call data that is now passed to audio_stream()


Dygear:
LGTM, nice job @aaknitt!

Issue #602, opened 2022-01-07
The rdio-scanner plugin does not send the freq_list with spike_count and error_count


chuot:
I just realized that the rdioscanner-plugin does not send the freq_list with the number of spikes and errors in the frequency field, so Rdio Scanner always shows "E: 0 S: 0" instead of the proper data.

I am however a little confused on how to bring this data to the uploader. What would be the right way to do this, add the information in `Call_Data_t`?

aaknitt:
Looks like when the plugin system was put in place that info didn't make it into the JSON in the same way that it used to be either:
https://github.com/robotastic/trunk-recorder/commit/4daac16a0cfea8bd34ad4d64db75abb12086e19f#diff-27578a9e34f9eb8cfcc62585171ceb71273ff44aa16f05c5f51e40322e58f2f3

So there is a freqList object available in the JSON file but it's not actually getting populated with all of the data from the freq_list.  Error Count and Spike Count look like they're hard-coded to zero and only one frequency is making it in there, the list isn't being accessed and looped through to populate the JSON data.  

I think you're correct that the freq_list information needs to get added to the Call_Data_t structure and then populated in call_concluder.cc.  The JSON generation should also loop through that same info to pull out everything from the list so that it's available there in addition to via the plugin system.

PR #635, opened 2022-02-09
Rdio scanner auto-populate / Better TG info for plugins

taclane:
The rdio scanner plugin has been attempting to auto-populate "Talkgroup Groups" using the variable `tg_group`.  The plugin currently is only provided the Tallkgroup ID# and Alpha Text, and cannot access other talkgroup information that may have been read in from the csv.

Complicating things, `tg_group` is presently creating downstream issues since it is a string variable assigned an integer without type conversion. This is resulting in non-printable characters and other garbage overwriting TG group assignments in Rdio-Scanner as seen below:
Screen_Shot_2022-02-07_at_9 25 26_AM

While the simple fix is to correct the type error, or `tg_group = "";`, the optimum solution would be to let plugins have all the talkgroup data loaded from the CSV to pass on as needed.

The call .json and other plugins could benefit by retaining this data as well.

The inability for plugins to have all the TG data was noted in the following rdio-scanner issues:
https://github.com/chuot/rdio-scanner/issues/123
https://github.com/chuot/rdio-scanner/issues/115

This PR works independently of https://github.com/robotastic/trunk-recorder/pull/630.

### DIFF
--- a/plugins/rdioscanner_uploader/rdioscanner_uploader.cc
+++ b/plugins/rdioscanner_uploader/rdioscanner_uploader.cc
@@ -45,8 +45,6 @@ public:
   int upload(Call_Data_t call_info) {
     std::string api_key;
     std::string system_id;
-    std::string tg_group;
-    std::string tg_tag;
     bool compress_wav = false;
     Talkgroup *tg;
     Rdio_Scanner_System *sys = get_system(call_info.short_name);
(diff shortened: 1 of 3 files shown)


taclane:
@chuot For your visibility since this affects the rdio-scanner plugin.

I've been running this for the last several days, and it has successfully been a way to pull TG info from the csv and automatically populate the fields in Rdio-Scanner.  ex:
```
5611,15eb,T,Conn DOT D1,District 1,Public Works,ConnDOT,1
```
to
Screen Shot 2022-02-09 at 12 09 27 PM

aaknitt:
I'm wondering if it would be useful to pass along the info for all patched talkgroups as well (rdio could then in theory display alpha info for everything in the patch).  

As an alternative to sending alpha tag and group info with each call, it may be possible to send all of that talkgroup info to rdio on startup using the setup_systems() function in the plugin (though .  It could then update everything in the rdio gui at once in case any of the alpha tags have changed, rather than having to wait for a call on that talkgroup to come through to get the new info.  Not sure if the rdio API would handle that as-is or not.

taclane:
While it can be done in the admin interface, I don't believe the rdio API presently has an endpoint to just upload a talkgroup.csv in bulk.

For practical reasons, I would be somewhat against pushing an entire talkgroup set by default anyway.  While the rdio interface is really nice, it can get bogged down with a large number of TGs, and it can be extremely cumbersome to use/manage when it has the full csv for a large system.  A couple hundred unnecessary buttons on the mobile app, or a seemingly-endless dropdown menu can make the user experience pretty rough.

Auto-populating talk groups as the calls come in at least guarantees that the TR instance is able to receive those TGs or rdio scanner may have recordings archived.  I am not sure the state (or future) of rdio-scanner auto-updating _existing_ TGs with updated info from TR.  At the moment, it is only supposed to work when populating new TGs, or updating TGs with empty fields such as "" or "-".

robotastic (changes requested):
Could you break this into 2 different PRs? The first to add additional Talkgroup fields to the Call_Info struct and the second to make changes to the Rdio plugin? I can approve the call_info ones, but I would want @chuot to review the changes to Rdio plugin

taclane:
Sure!  I'll split the PR as soon as I get a free moment.  

It would be great to get some feedback from the Rdio Scanner side, but core of this is just to connect the dots from information currently in the trunk recorder talkgroup.csv to the RS API endpoint which already accepts `talkgroup`, `talkgroupLabel`, `talkgroupName`, `talkgroupGroup`, and `talkgroupTag` as inputs.

https://github.com/chuot/rdio-scanner/blob/1eebbc115432915bf7f45473fd7e3219005e4a6c/server/parsers.go#L268-L291

chuot:
@taclane Sorry for being late. It's been a crazy week and things have piled up.

First of all, I want to thank you for handling this issue. Everything looks clean and compliant with the Rdio Scanner API.

Would the new call_info structure allow us to retrieve the number of errors and spikes that were missing from v4 (issue #602)?

@robotastic Changes approved

taclane:
@robotastic It looks like chuot is okay with the changes to the plugin, but let me know if you'd still prefer I break this apart. 

@chuot This PR alone shouldn't be able to pull the spikes and error counts out of the `freq_list`.  I can look into that next and see if there's any particular reason they've been hard coded to zero in the plugins/json.

chuot:
@taclane Thanks a lot!

robotastic:
@taclane Could you check out the `tg-info` plugin? I moved it around so the string from the TG are copied into the call-info struct:
https://github.com/robotastic/trunk-recorder/tree/tg-info
I also made the same changes you had for the rdio plugin. If things look good, I will merge that branch in.

taclane:
@robotastic 
It looks like it's working great!  I deleted all the talkgroups in rdio-scanner, and with each call it's repopulating [ID, Label, Name, Group, Tag] with info straight out of the talkgroug.csv in trunk-recorder.

robotastic:
Great - I will merge that branch in! If everything looks good, then we can cancel this PR.

Issue #628, opened 2022-02-01
Clarification on Patch Handling


blantonl:
Documentation is missing on how patch handling is occurring for P25 systems.  

Can the team advise or update the docs?

aaknitt:
Externally, there's a new field in the JSON data that's associated with the audio file called "patched talkgroups".  When a patch exists, it will be an array of all talkgroups that are part of the patch (including the one already listed in the "talkgroup" field).  Here's an example:
```
{
"freq": 852612500,
"start_time": 1644029614,
"stop_time": 1644029620,
"emergency": 0,
"encrypted": 0,
"call_length": 4,
"talkgroup": 58916,
"talkgroup_tag": "WKLEDISP3",
"audio_type": "digital",
"short_name": "oasiswauk",
"patched_talkgroups": [58916,58918],
"freqList": [{ "freq": 852612500, "time": 1644029614, "pos": 0.0, "len": 4, "error_count": 0.0, "spike_count": 0.0}],
"srcList": [ {"src": 6815080, "time": 1644029614, "pos": 0, "emergency": 0, "signal_system": "", "tag": ""}, {"src": 6815080, "time": 1644029618, "pos": 3, "emergency": 0, "signal_system": "", "tag": ""}, {"src": 6800006, "time": 1644029619, "pos": 3, "emergency": 0, "signal_system": "", "tag": ""} ]
}
```
Internally, patches are tracked in the System object.  The System ->get_talkgroup_patch() function can be called to get a list of talkgroups that are patched with the talkgroup ID that's passed in to the function.  

```
std::vector<unsigned long> get_talkgroup_patch(unsigned long talkgroup);
```

An example of how a plugin can access this info can be found [here](https://github.com/robotastic/trunk-recorder/blob/12e664105b698ad2b51a96b73c68afa0fc093e25/plugins/simplestream/simplestream.cc).  

History on the changes to add the feature can be found [here](https://github.com/robotastic/trunk-recorder/issues/368).

PR #637, opened 2022-02-19
Add support for M/A-Com (Harris) P25 patch following

aaknitt:
Per ongoing discussion [here](https://github.com/robotastic/trunk-recorder/issues/368) and [here](https://github.com/robotastic/trunk-recorder/issues/628)


robotastic:
Nice work! Thanks @aaknitt & @blantonl for getting this figured out

Issue #646, opened 2022-03-05
Allow simplestream to send directly to pulseaudio


aaknitt:
Putting this here for tracking, I'll look into doing it, but open to input.

Currently a "translator script" is needed to receive UDP audio from simplestream and feed it to pulseaudio.

pulseaudio can receive audio directly via TCP or via RTP.  Being able to send directly from simplestream to pulseaudio using one of these two methods could be beneficial.  

RTP:
https://wiki.archlinux.org/title/PulseAudio/Examples#Using_RTP/UDP_instead_of_native-protocol-tcp
https://en.wikipedia.org/wiki/RTP_payload_formats#cite_note-1

TCP:
module-simple-protocol-tcp
https://www.freedesktop.org/wiki/Software/PulseAudio/Documentation/User/Modules/

Enabling TCP should be simple and require least amount of changes.  RTP may be more useful/flexible long-term (could be used by more than just pulseaudio).

aaknitt:
TCP is implemented in #649 .  Closing for now, can be reopened if there's a strong desire for RTP at some point.

PR #649, opened 2022-03-12
Add TCP support to simplestream plugin

aaknitt:


aaknitt:
Related to https://github.com/robotastic/trunk-recorder/issues/646

Issue #651, opened 2022-03-18
Stopping recording when talk group changes on voice channel


rosecitytransit:
A while back, I think they changed the configuration on the radio system I monitor and now there's sometimes very little time between calls on the same frequency.  As a result, both calls sometimes recorded into one file (the second call also correctly gets its own recording). I made [a waterfall heatmap](http://www.rosecitytransit.org/files/heatmaps/20220309-SDR01-trimet-hour-anno.png) using heatmap.py and highlighted a couple examples of it happening. The numbers correspond to calls [here](http://www.rosecitytransit.org/radio/?All+3024+2022-3-9).

Note that I keep calls open through the post-voice TDU, etc frames because I prefer to keep conversations together rather than record by transmission (so it's easier to find interesting stuff) and since the unit-unit calls don't correctly continuously show on the control channel like they probably should.

I see there's #564 and #626 but I'm wondering if it would instead be possible to stop recording based on checking the talk group in the voice channel metadata.

Specifically, I'm wondering if it would be possible to get the talk group from the call into p25p1_fdma.cc to compare with `grpaddr` and block processing if they don't match (e.g. prevent the `process_voice(A);` that happens right after that `process_LCW` section happens. https://github.com/robotastic/trunk-recorder/blob/5d3da98d39c2a96e27e4825c0080fc83e8c8b5c7/lib/op25_repeater/lib/p25p1_fdma.cc#L464

I know one option would probably be to add a talk group field to `rx_status` like I added `last_update`.

Also, the ideal would be that a call with the new talk group is searched for/set up, in case a control channel message for it is missed.

Thanks!

robotastic:
I would check out the current version of TR and maybe try the `no-overlap` branch. The current approach is to use TDUs to separate the transmissions, but keep them joined together in a conversation. I tried also to use the SRC / GRP ID from the voice channel, but it is either sent out weird, decoded wrong on SmartNET digital channels, and it resulted in transmissions started and stopping weird places.

The call control has changed a bit too. The block that recorder wav files now handles keeping track of each of the transmissions. The same info can be passed along as TAGs, the way src_id and TDUs are being handled.

rosecitytransit:
Thanks for responding! It is true that I haven't touched things in quite a while and am still using v3.3, though I am starting to play with the current version and am willing to upgrade.

[...]

PR #653, opened 2022-03-22
Framework to handle U2U Calls and config option to ignore them.

tadscottsmith:
This is about as far as my testing is stable. I continue to troubleshoot, but I am struggling to get more than one side of the communication. I've tested a few other P25 monitoring applications and they're behaving the same. I actually think this might be because the only agency that uses U2U is a shared mobility ride service for those with disabilities and health conditions. My gut is that I'm hearing the operators who are on a system I'm monitoring, but not hearing the drivers who are roaming all over the metro on other systems I'm not monitoring.


robotastic:
Oh interesting - are there multiple sites, so the call wouldn't be transmitted on all of them? It is very possible that the P25_parser is missing some of the messages for UU calls too.

tadscottsmith:
Yeah, the dispatchers are centrally located, but the I imagine the van drivers could be on any of up to 15 different sites within the same system. Have you run into issues before where the Multiple Block Trunking (MBT) messages aren't being processed? That's what I'm currently debugging, because based on the specs, it seems the TSBK is only applicable when _the target and source unit are members of the same HOME system and currently are residing at this HOME system, AND the channel identifier implicitly indicates both the receive and transmit frequency values for the assigned channel._

rosecitytransit:
My guess is that what's happening is that the dispatch radio(s) are affiliated the site that you are monitoring, but that the operators radios are not.  Like when no radio on a particular talk group is affiliated with a particular site, you will only hear the sides of conversations that need to be broadcast to reach a particular radio. I know I have one-sided conversations of unit calls on a SmartNet system when one party was not on the same site.

Or that the system is configured not to rebroadcast the mobile side, with the idea that the dispatchers would be getting the audio through a backbone connection. (Kind of like how taxis would use two separate frequencies so drivers couldn't hear other drivers.)

tadscottsmith:
I think this is probably the case. Return audio from the drivers to the dispatchers is being backhauled from a remote site to the local site over the network and the dispatchers are at some sort of network connected console. The weird thing is that the dispatcher traffic is being aired over the air on the local site, and not solely backhauled to the remote site. I'll try to grab some trace files tomorrow since the only agency that uses U2U on my system is mostly an 8-5 operation.

rosecitytransit:
I should add that, while I don't know how the system you're monitoring is set up or how the calls are appearing, but on the one I monitor all unit calls come from consoles with a specific range of radio IDs, and if you want to block them you should be able to just enter them in as talk groups and block those, e.g. by setting the priority way high.

tadscottsmith:
This PR actually adds a config option to ignore them globally. They have different grant/update message types, so U2U calls should be pretty easy to discern. My primary concern is that on this no-overlap branch, there might need to be some work done to accommodate for working U2U calls. Since the TG flips back and forth between the original sender / receiver, that needs to be accounted for. I err on the side of caution since I don't have a system I can test against that had working U2U calls on the master branch.

rosecitytransit:
re the no-overlap branch, one question is whether the "flipping" should be considered a new transmission (like it was two people talking on a group channel) or whether the complete calls should always be kept together as a single conversation. Also, I'm having trouble getting my workaround for dealing with how the unit calls don't show up continuously on the control channel on the system I monitor (see https://github.com/robotastic/trunk-recorder/issues/651#issuecomment-1081401260 )

tadscottsmith:
I'm going to close this PR and create a new one based on all of the latest changes to the master branch.

PR #654, opened 2022-03-23
Add per-system configuration option for minimum transmission length.

tadscottsmith:
Add a minTransmissionDuration per-system configuration option. This represents the minimum transmission duration in seconds (decimals allowed). Transmissions below this number will not be added to their corresponding call. The default is 0 to ensure that no transmissions are allowed without updating the configuration.


robotastic:
I like it! .... but I am going to switch it around so that call concluder handles this. I want to make sure that the small files get deleted or else they might just clutter up a system if they are set to delete.

PR #655, opened 2022-03-25
GNU Radio typo

777arc:
<3


robotastic:
I will be very impressed if this is the only place I had that typo! 🚀

PR #659, opened 2022-03-31
Fix op25 errors

tadscottsmith:
This seems to be the source of the _op25_audio::hostname_to_ip() getaddrinfo: Name or service not known_ errors on startup that started in November. The commit after the one below lost the the default "127.0.0.1" value.

https://github.com/robotastic/trunk-recorder/blob/fa72e5879328646f46d57ad537320788045d7c16/trunk-recorder/recorders/p25_recorder_decode.cc


robotastic:
Nice catch!

rosecitytransit:
I believe these options should really be exposed and settable in the configuration file. I know the ```verbosity``` option turns on OP25 debugging (which ideally would be switched to BOOST logging), and I think the UDP options are for theoretically sending the traffic to Wireshark (at least if the P25 plugin for it was current)

Issue #525, opened 2021-10-08
unitScript functionality doesn't seem to be working in 4.0-beta


aaknitt:
unitScripts specified in config.json that fire and function correctly in 3.3.7 don't appear to fire in 4.0-beta.

config.json:
```
{
	"ver": 2,
    "sources": [{
        "center": 853175000,
		"ppm":0,
        "rate": 1400000,
        "gain": 40,
        "digitalRecorders":2,
        "driver": "osmosdr",
		"device": "rtl=0"
    }],
    "systems": [{
        "control_channels": [853712500,853687500,853350000,853137500],
        "type": "p25",
		"modulation": "qpsk",
		"digitalLevels":8,
		"shortName": "oasiswauk",
		"talkgroupsFile": "OASISTG.csv",
		"unitScript":"unitscript.py",
		"recordUnknown": false,
		"audioArchive": true,
		"talkgroupDisplayFormat": "id_tag"
    }],
	"minDuration": 0,
	"defaultMode": "digital",
	"captureDir": "/home/pi/tr-recordings",
	"controlWarnRate":10,
	"controlWarnUpdate":1,
	"frequencyFormat": "mhz",
	"logFile": true,
	"callTimeout": 2,
	"statusAsString": true
}
```
minimal test script unitscript.py:
```
#!/usr/bin/python3
print("THIS IS A UNIT SCRIPT")
```
Results in 3.3.7:
![image](https://user-images.githubusercontent.com/13441553/136576208-b90fa108-508b-4cf0-a6f3-77296c3e7a33.png)

No evidence of script being run in 4.0-beta:
![image](https://user-images.githubusercontent.com/13441553/136576384-4096b9f3-e46c-41b4-950a-e590072e214d.png)

robotastic:
Ok - this commit ba6959fc655cc70272506e7cd9e62f2d7798c8fc should fix the unitScript issues. Let me know if that doesn't work.

rosecitytransit:
I know this is way later (I've just been letting t-r run), but the "./" should be in your config.json entry as `"unitScript":"./unitscript.py",` I thought I noted that in the documentation I wrote; it's just like if you were to execute the script on the command line.. `./` would be incorrect if the script is somewhere else besides the same directory as trunk-recorder.

PR #661, opened 2022-04-03
Log data channel grants, unit answer requests, Location Reg Responses

rosecitytransit:
-Decode and log tsbk 0x05 Unit to Unit Answer Requests and 0x14 SNDCP Data Channel Grants
-watch for different talk groups in a call on the voice channel (though I think the current master may negate the need for this; also I didn't actually stop the recording, just log it happening)
-update CONFIGURE.md unitScript entry
-undo #525; the issue is that the config.json entry needs to include a ```./``` if needed; it shouldn't be hard coded as if you want to run a program, there should be no ```./``` at all.

If you have issues with any of those, I can easily separate them out. Also, still want to put together/submit the ability to adjust the control channel decoding rate interval; see https://github.com/rosecitytransit/trunk-recorder/commit/02969eed165fa259ee0bca990af13b35bf5d3b5a and https://github.com/rosecitytransit/trunk-recorder/commit/59c89a1de8be2f90812a904808785a5a0825997f and https://github.com/rosecitytransit/trunk-recorder/commit/25376e774cdcbe292dcdfb6f7797953f91c46367 Also might add the option to put the JSON data in a single, daily file instead of one for each call (if you're not uploading)


robotastic:
Cool - It would be helpful if you could break these up into separate requests. I am also about to merge in the `no-overlap` branch. Could you rebase off of that branch or just wait a few days until that is merged in?  I don't think there should be any big changes, but don't want to create any churn.

I was thinking about adding in a Group check on the voice channel. I was going to send the Group down to the wav-sink using tags, similar to the way that SRC ID is done. All of the intelligence about a Transmission is living there now. I want to keep the OP25 code as close to the baseline as possible and also try not to put much intelligence there. I can work on implementing this approach soon.

For alternate JSON files, it is best to do that as a plugin... that said, I really want to find a way to make it easier to have standalone plugins outside of the repo, so people can easily update them without me having to approve changes.

rosecitytransit:
Even though I've wasted way too much time trying to scratch the itch of getting the voice channel checking to work, I'll leave that out. I am hoping that it's not needed with the current version of the code.  I haven't come across the issue of extraneous audio in the testing I've done so far; my hope is to update the running copy tonight.

The other changes are pretty basic, just decoding and logging a few more trunking messages. As for the JSON thing, I'll just do that in my fork. Except for the overhead of having the individual files be written and then parsed, one could use the encoding script to get things how they want.

Regarding the no-overlap branch, I tried it, but I unfortunately need the ```rx_status.last_update``` code to keep unit-unit calls open for recording. The master branch works fine. Plus, so far it doesn't seem to be needed to address overlap/extraneous audio issues.

I really wish GitHub would allow you to submit patch files, instead of having to submit a copy of the entire code, though I am learning more about branching.

robotastic:
Looks Good - thanks for whittling this down.

PR #662, opened 2022-04-03
Set alpha tag for conventional talkgroup and restore startup console message

taclane:
Conventional channels using a .csv currently do not have their alpha tag displayed in the trunk-recorder console messages.  As in previous releases, the alpha tag should be when the conventional channel is started.

Also to be consistent with previous releases, the "Monitoring Conventional Channel" startup message should display whether or not a conventional csv is being used.


robotastic:
Good call...

PR #663, opened 2022-04-06
minor fix to unit answer reqest logging

rosecitytransit:
looking at the merge commit, I saw that only one side's radio ID was being logged

Also, if you want, I can maybe add control channel decoding rate warning rate config ( https://github.com/robotastic/trunk-recorder/pull/470/commits/2d6a95484f3919a9ff4af4cde92db315d847ff04 but I have it separated from the actual decoding rate check) to this PR


robotastic:
This makes sense - good catch.

That submission for control channel warn update rate also makes sense - could you add to the config.md document describing what this does?

rosecitytransit:
I'll definitely do so when I submit it. I need to put it together so it keeps the actual decoding rate checking at 3 seconds.

PR #666, opened 2022-04-11
add handle_message plugin hook, edit README.md

rosecitytransit:
Control how often the Control Channel Decoding Rate is logged; sometimes you want it every 3 seconds to test things, other times once a minute suffices. Keeps the actual rate checking at 3 second intervals.

Originally https://github.com/robotastic/trunk-recorder/commit/2d6a95484f3919a9ff4af4cde92db315d847ff04


robotastic:
I am a little confused on this approach - it looks like this changes it so it will only perform the check at a configurable interval now, right? The challenge with this is if you have very bursty changes you would miss short interval drops. Like it would be checking a 3 sec interval every 30 sec.

What is the overall goal? With the threshold, it should only print this messages when things are bad and this shouldn't be too often. Are you looking to instead occasionally have an output of the channel decode even when things are above the threshold?

rosecitytransit:
It's supposed to leave the internal checking (to make sure control channel messages are still coming in) at 3 seconds but only output the "Control Channel Message Decode Rate" messages to the log file if it's been X seconds since last doing it (e.g. if set to 60, only output every 20 checks) . The goal is to reduce the size of the log files a bit yet still log the rate to see how decoding is going.

rosecitytransit:
>  Are you looking to instead occasionally have an output of the channel decode even when things are above the threshold?

Correct, I use the data for the decoding rate graph and to confirm every thing is working OK

rosecitytransit:
@robotastic you want to merge this in?

robotastic:
> And is there a reason why the iteration at https://github.com/robotastic/trunk-recorder/blob/master/trunk-recorder/main.cc#L1499 can't be combined with the iteration in check_message_count at https://github.com/robotastic/trunk-recorder/blob/master/trunk-recorder/main.cc#L1379 ?

Yea - it might be possible to combine the 2, or move the Stale Patch checking to its own function, but lets leave that alone for now.

I think there maybe 3 different values at play - how often we would want to display the warning message, the period of time over which the average should be calculated and the message rate value that triggers if the message should be displayed.

There are probably 2 different scenarios where this would be used:
- I want to get an error message right away if my message rate drops below a value and I don't want the average to be over a long time because it might miss spikes.
- I want to occasionally get the decode rate so I can keep an eye on things. Averaging period might not matter as much here.

I sort of arbitrarily picked 3 seconds as the period of time to average over.... not sure that is right, but maybe it is good enough. 

It would be nice to support both use cases at once, print the message every x seconds if the rate is good and print it right away if it is bad. 

Let me see if I can rework things to support that

rosecitytransit:
Thanks for getting back to this.

> I want to get an error message right away if my message rate drops below a value

What if I change `if (msgs_decoded_per_second < config.control_message_warn_rate || config.control_message_warn_rate == -1) {` instead to `if (msgs_decoded_per_second < config.control_message_warn_rate || (config.control_message_warn_rate == -1 && warn == true) {`? I think this would log the rate every 3 second interval if it's below the controlWarnRate threshold AND always log it at the controlWarnUpdate setting. The second half could be just `|| (warn == true)` or maybe `|| (config.control_message_warn_rate == -1) || (warn == true)` by differentiating between an explicitly set controlWarnUpdate and the default.

> I sort of arbitrarily picked 3 seconds as the period of time to average over.... not sure that is right, but maybe it is good enough.

Well, the 3 seconds seems to come from how long the program wants a lack of messages for to decide to retune. I think it's a fine amount.

tadscottsmith:
Since it's an or `||` in line 1410, won't this still print out low message decode rates regardless of the value of `print_message_rate`? 

Additionally, I'd suggest `controlMessageUpdate` default to `-1` so that the behavior doesn't change unless someone specifically adds a `controlMessageUpdate` option.

Finally, this seems like something that should be handled outside of the trunk-recorder logs... I don't understand what the use case would be for expecting bad decode rates and not writing an error to the console? If decode rates are that low, a number of other issues are going to pop up and be even harder to troubleshoot.

robotastic:
@tadscottsmith the behavior I was aiming for was to print out the current decode rate whenever it falls below the threshold and also every 30 seconds or so... did I mess that up? I have been doing JS for the past month so my C++ is rusty.

You are probably right though that instead of defaulting to 30 seconds, it should be -1 and people can optionally enable it.

The target use for this would be spotting when the decode is low, but is working well enough to capture most of the system. Could be a helpful signal that you need to adjust something... but fair question - maybe it would be better to have this be an MQTT thing

tadscottsmith:
My two cents and my two cents only:

I really would be against printing an "error" level message if things are properly working. I have gotten to the point where I go days and hundreds of thousands of calls without an error, and I think that's a good thing.

Do we really need something more granular in the logs than when the print_status() runs? The plugin system already gets the messages every 3 seconds.

rosecitytransit:
I do think it may be a good idea not to change the defaults.

> did I mess that up?

Line 1410 looks OK to me. It's a simple if statement. Note that some may set the decoding threshold (controlWarnRate) high, and want logging when there's even a modest decrease.

> I really would be against printing an "error" level message if things are properly working.

I'm guessing that the idea is to always print the message even if logLevel is set to `error,` e.g. if one just wants to make sure every thing is working and not any details. `info` might be more technically appropriate.

> The plugin system already gets the messages every 3 seconds.

If the plugin system gets it, and there's a way to save the rate to a file, I'd be fine just using that and turning off the messages in the logs completely. I'd just have to modify the decode rate chart's data source.

tadscottsmith:
Would it make more sense to add a section in print_status() that prints the current decode rate for each system, and then make the timing of how often print_status() runs a user option?

rosecitytransit:
> Would it make more sense to add a section in print_status() that prints the current decode rate for each system, and then make the timing of how often print_status() runs a user option?

Maybe we should just do that and then only print the message every 3 seconds if it's below controlWarnRate. Would simplify things, add flexibility and maybe help with debugging. The only downside is that (especially if you have a ton of recorders) it would mean more log file lines if doing it often, but I'd be willing to just address that in my copy (if I can't just use a plugin).

rosecitytransit:
I was going to offer up to code it if there was consensus. My two thoughts are just:

*if it's print_status() time, don't output the decoding rate if it's low since the print_status() is going to do it

*it should maybe be error level if the decoding rate is below the threshold, otherwise info like the current the status message.

> I have gotten to the point where I go days and hundreds of thousands of calls without an error, and I think that's a good thing.

I don't ever really look at the log fines unless something has happened (e.g. issue w/control channel change. which is being addressed in 792)

Lastly, I'm not seeing a plugin to get the decoding rate

robotastic:
Ok - I made the change so it is part of the Status message that gets printed out. I am going to leave it at the current fixed delay - which is 200 seconds. No one has asked for it to be more or less ever or to configure it, so let's see how it looks. Check the `rate-status` branch

rosecitytransit:
Being able to control the status updates might be a nice to have. As I said, I'm just wanting a way to get the rates for the decode rate chart yet not have them be logged 20 times (60/3) more than necessary (though the chart could easily be changed to handle more than 1 point per minute).

Lastly, if you haven't noticed, I also included a config entry for the logDir option.

rosecitytransit:
If you want to leave things as-is; I'd be willing to just customize things on my end and just submit the logDir config entry (unless you want to write that, in which case I could close this).

one other minor, somewhat related issue: I think the timestamp for `last_decode_rate_check` should be first set at 0 here: https://github.com/robotastic/trunk-recorder/blob/7bb5742a88d816d4ce3f8a68053d70e2e9cffbc8/trunk-recorder/main.cc#L1433

Now, there's a delay at startup because the program waits for the 3 seconds of no control channel messages to do a retune.

rosecitytransit:
Well, although it meant another day delay in getting things upgraded, I've decided to switch to logging the rates in a separate file. See https://github.com/robotastic/trunk-recorder/commit/0cd6c66c240abf2fd20017bea6e36991950d4695

While it means another file, it will mean not having to grep through what can be very large log files, not needing to keep (or even have) log files, and save size vs the multiple "Control Channel Message Decode Rate:" messages per cycle. Need to still update the decoding chart.

I can change this PR to that method if this is desirable. I now realize there is a (undocumented?) `plugman_system_rates(systems, timeDiff);` method to put it in a plugin; I like the efficiency using the existing for loop. Also, while I did use the exiting `last_status_time` variable and turn off the status message printing (personal preference now), creating another variable is a very simple thing.

taclane:
> I now realize there is a (undocumented?) `plugman_system_rates(systems, timeDiff);` method to put it in a plugin; I like the efficiency using the existing for loop.

Yup. There's a lot of power in the plugin architecture to solve exactly something like this.  Here is an example usage in the MQTT plugin

https://github.com/robotastic/trunk-recorder-mqtt-status/blob/d37d02ea131dc758f079ee88181bcb2e42a47a48/mqtt_status_plugin.cc#L85

but it's likely pretty simple to build a single-purpose plugin to only output `system_rates` wherever and however frequently one might want.

robotastic:
Sorry - I have been over focused on trying to debug why my OpenMHz code isn't working. 

I don't see the `logDir` config as part of this PR. Would you create a separate PR for that?

The creation of a log file for building a decode rate chart is best handled by a plugin. Let me know if there are any kind of  changes required to the plugin system to help support it. I think everything should be there...

For the `last_decode_rate_check` wouldn't you have to wait 3 seconds anyhow to accumulate enough messages to see if you can receive things well enough or not?

rosecitytransit:
No problem and thanks, as I've noticed some wonkyness with the site. As for logDir, see https://github.com/robotastic/trunk-recorder/pull/666/commits/bc23e8805c388c7c6eb557706a6d18151f098782 I can put it in another PR if we're unsure about this one. The actual functionality should already exist.

As for decode rate logging, would you like me to switch this PR to using the method in https://github.com/robotastic/trunk-recorder/commit/0cd6c66c240abf2fd20017bea6e36991950d4695 , and then someone (maybe me) can move it to use a plugin?

> Yup. There's a lot of power in the plugin architecture to solve exactly something like this.

If it was me I'd just have a lot of places where it could optionally run a script with certain variables, but I agree that having access to the actual internal data is useful, and adding code in a separate plugin file prevents potential future conflicts.

rosecitytransit:
This is now a plugin that logs the decoding rates to decode_rates.csv in the capture_dir. A header line is put in the file each time the program started. This only logs the rate over the last 3 seconds; I am thinking of having it calculate the rates over the entire interval by saving and adding to the numbers each time the plugin is called.

rosecitytransit:
@robotastic this should be good to go, except for the rates being for only the proceeding 3 seconds. I know I could create a vector in the plugin to save the rates if it's not time to log yet, but I'd like a separate counter (or maybe plugin call) comparable to `sys->set_message_count()` https://github.com/robotastic/trunk-recorder/blob/e33e441278ae130a6884d464a5d75b97655ca5fd/trunk-recorder/main.cc#L1480 so it could be independent of `check_message_count()` and log at any chosen interval.

I was thinking of maybe creating a plugin call in `handle_message()` https://github.com/robotastic/trunk-recorder/blob/e33e441278ae130a6884d464a5d75b97655ca5fd/trunk-recorder/main.cc#L1228 which could be useful for other ways to handle trunking messages. Your thoughts?

Also, ignore what I said about the `last_decode_rate_check`; on closer inspection it's not waiting for the time_diff at startup like I thought it was.

rosecitytransit:
Thinking about and looking at things, how about I (maybe in a separate issue, before this one gets addressed) create a `handle_message` plugin that passes the message and system and then move the following 2 sections that are only used for unit script to that plugin? I'm guessing things could be simplified by having one function with an if statement to set the actual text sent. I know this would be a change, but it would move the stuff out of trunk-recorder, the non-call plugins aren't documented anyways, and I don't think there's going to be a lot of demand for one dedicated to e.g. acknowledge responses.
https://github.com/robotastic/trunk-recorder/blob/592e551f9f7153986ab2f439e4bbc3e0f54e8a38/trunk-recorder/main.cc#L1257-L1300
https://github.com/robotastic/trunk-recorder/blob/592e551f9f7153986ab2f439e4bbc3e0f54e8a38/trunk-recorder/main.cc#L964-L996

rosecitytransit:
> It makes it obvious for people writing plugins that those message types exist.

I would just have documentation that they do, or point people to the plugin.

> Over on the Discord server

I know I should give in and create an account and go through all of/join the discussion there (why doesn't GitHub have a chat function? Or at least a discussion section just like "Issues" but not for specific bugs/feature requests/etc.)

> depending on what you are doing it might be easier to just have a switch statement in your plugin instead of a lot of functions.

Exactly. I said "if statement" but switch would be more appropriate if handling a lot of types.

> Good point - this call could also be used for just counting the number of messages.

Which is just what I want to do. :)

> Having a separate repo keeps the build times down and let you make updates without having to do a PR.

As long as there's simple, explicit instructions on how to add external plugins (how to merge the code in). I'll modify this PR to just be a `handle_message` plugin hook and the changes to README.md I made.

taclane:
Just to check in from #TeamDiscord, we could probably go either way on this one.  Both the current switch/case tree or a monolithic `handle_message`-type thing would ultimately work the same; the latter just offloads the work to every plugin to reimplement it's own switch/case.

Though, the advantage of the present system is that it essentially allows plugins to "opt in" to messages of interest.  If a plugin doesn't want to get woken up for every `ans_req` or `ackresp`, it simply omits a handler for that message.

> As long as there's simple, explicit instructions on how to add external plugins (how to merge the code in). 

Check out some of the example plugins that have been posted above.  The nice thing about the external plugin system is that they can be developed, compiled, and installed entirely separate from trunk-recorder without having to modify the upstream code.  The only thing external plugins require are the headers and libraries t-r puts in place during a `make install`.  A lot of the territory is undocumented, and we're testing the limits of what they can do, but it is a simple and powerful way to tweak trunk-recorder to suit personal data collection or reporting needs.

I'm hoping to reconcile a few things before I push the community fork back to the original, but if you want to check out an example of an external plugin that uses just about all the API hooks without changing any of t-r's source files, [this is what we've got so far](https://github.com/taclane/trunk-recorder-mqtt-status).

rosecitytransit:
It's true that `handle_message` gets called constantly (up to 26x or 40x/second/system for P25) but I think most plugins would either use the separate call_start or setup_system hooks or want to know about many types of messages. For the unit_script and similar plugins, one idea is a hybrid approach where there's a single hook that gets called for any type of unit-involving message. That is, this would be squashed into a `void unit_message(TrunkMessage message, System *sys)`: https://github.com/robotastic/trunk-recorder/blob/0dbe2f3e3974b3e959394452f7fec808e7d0f470/trunk-recorder/main.cc#L970-L996

> Check out some of the example plugins that have been posted above.

Have done that somewhat and think I am going to do that for the decoding rate plugin, and could for the unit_script one.

> The only thing external plugins require are the headers and libraries t-r puts in place during a make install

Actually no. If you're not installing, you just need to add the 2 lines I added links to in my last comment and put the plugin in the plugins folder (or wherever) . Unless my grep of "unit_script" is inconclusive.

For the moment, I'm working on switching to the new_call_mgmt branch because my approach in https://github.com/rosecitytransit/trunk-recorder/tree/fe20271d9676a5874e5d08dcb04d1fb51dc2de22 isn't working as desired.

robotastic:
I am happy to take a belt and suspenders type of approach:
 - Let's keep the existing Plugin hooks the way they are. I think it is better to  have a 1:1 relationship between hooks and messages to keep it less confusing. It doesn't use much CPU comparatively and doesn't take up much code space.
 **ACTION**: make sure all Hooks are documented
 - Let's add a new hook that pass all of trunk messages to a plugin
 **ACTION**: add new hook
 
 Plugins that are not part of the TR repo should be in different directory because they going to be in a separate directory. They will have to be installed because the plugins could be in any random directory.

taclane:
That sounds great!  Both approaches have their uses cases. The thing I still like about the current setup is that it allows plugins to be lightweight if their required inputs are minimal. 

For example, a theoretical plugin to log the metadata of encrypted calls would only have one API hook to worry about instead of parsing each and every trunking message.

rosecitytransit:
> because they going to be in a separate directory

You mean a different repository and how that could mess up git? If so, I think you could have a second, "external_plugins" directory that's put in trunk-recorder's .gitignore file, and then have the base of the plugin repository be the plugin's own directory. Or just tell people to manually add/manage the directory.

As for the 2 lines, what about a sed script for people to run that does e.g. `sed -i s|//PLUGIN LIST|//PLUGIN LIST\nadd_plugin|`?

> Let's add a new hook that pass all of trunk messages to a plugin

Are you going to do that, or do you want me to use this PR to do so?

> a theoretical plugin to log the metadata of encrypted calls would only have one API hook

Couldn't that use the existing call_start hook, or maybe the signal one? It appears that it applies to all known calls, including ones that just get monitored and not only ones being recorded. I definitely think it's worthwhile having one for calls (and being able to rely on trunk-recorder's call-tracking logic), i was just proposing combining the ones about unit messages.

robotastic:
The way it is setup right now, you just have to compile/install an external plugin and then point to it in the config. Let's move any alternate approaches to a new Issue. 

I just created a new branch that adds a trunk-message hook:
https://github.com/robotastic/trunk-recorder/tree/feature/message-hook

Give it it a check and I will merge it in

taclane:
> The way it is setup right now, you just have to compile/install an external plugin and then point to it in the config. Let's move any alternate approaches to a new Issue.
> 
> I just created a new branch that adds a trunk-message hook: https://github.com/robotastic/trunk-recorder/tree/feature/message-hook
> 
> Give it it a check and I will merge it in

Cool, thanks!  I'll check it out this week.  

Pushing everything through one API call can also accomplish two things that I have seen people ask for on discord (or gitter) before:
1. Is there a way to see all the trunking messages (besides turning on trace/debug logging)?
2. If there is a trunking message I'm interested in but isn't being decoded, how do I get it out of the parser?

robotastic:
Handle in PR #814 - Closing

Let's move any further discussion to a new PR.

rosecitytransit:
> Let's move any further discussion to a new PR.

No problem. Do you still want a PR with just a link to a decode rate logging plugin? Should I just add it to the readme where links to the other external plugins are?

robotastic:
That would be great! Yep, add link in the Readme and submit a PR

PR #668, opened 2022-04-14
Fix missing audio on rdio-scanner web and mobile devices

kb2ear:
Was having issues with rdio-scanner not playing all audio files, this mod seems to have fixed the issue.


robotastic:
Interesting!! It looks like this could work

PR #675, opened 2022-04-25
Fix short transmission deletes.

tadscottsmith:
Simply continuing in the **create_call_data** function is causing later functions (**upload_call_worker**, **combine_wav**) to have issues when they try to iterate the transmissions_list and the files have been removed. This should remove the entry from the transmission list and correct the issue.


robotastic:
Great spot!! I just implemented this fix but did it in a slightly different way to copy the pattern I had been using to delete while in a loop. I didn't switch the OP25 decoder yet. Probably should be a different PR with some discussion.

robotastic:
Let me know if my latest commit fixes things and I can close this PR

tadscottsmith:
Your latest commit seems to handle this correctly, thank you! Sorry about the OP25 decoder, not sure how that drifted into this PR. I'll close this. Thanks again!

PR #680, opened 2022-05-07
Update macOS Build Instructions

dechilders:
Added build instructions for gr-osmosdr since it is no longer available from Homebrew.
Added pybind11 as a dependency for building gr-osmosdr.

### DIFF
--- a/docs/INSTALL-MAC.md
+++ b/docs/INSTALL-MAC.md
@@ -2,6 +2,19 @@
 There are two main "package managers" used on MacOS: Homebrew and MacPorts. Trunk-recorder can be installed with dependencies from one or the other
 
 ## Using Homebrew
+Tested on macOS Catalina 10.15.7 with the following packages:
+
+- homebrew 3.4.10
+- cmake 3.23.1
+- gnuradio 3.9.3.0
+- uhd 4.2.0.0
+- pkgconfig 0.29.2
+- cppunit 1.15.1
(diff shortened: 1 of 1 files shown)


robotastic:
Ah Good catch! Thanks for making the update

e850205:
I had to use a different path for openssl to build on Mac - /opt/homebrew/Cellar/openssl@3/3.0.3

So the command for me was:
cmake ../source -DOPENSSL_ROOT_DIR=/opt/homebrew/Cellar/openssl@3/3.0.3

e850205:
Can't make on Mac right now anyways

"In file included from /Users/**/trunk-recorder/source/trunk-recorder/systems/system_impl.cc:1:
/Users/**/trunk-recorder/source/trunk-recorder/systems/system.h:40:59: error: unknown type name 'ß'"

taclane:
> I had to use a different path for openssl to build on Mac - /opt/homebrew/Cellar/openssl@3/3.0.3
> 
> So the command for me was: cmake ../source -DOPENSSL_ROOT_DIR=/opt/homebrew/Cellar/openssl@3/3.0.3

Homebrew installs in `/opt/homebrew` on apple silicon, and `/usr/local` on intel.  
To keep both bases covered, the cmake command using homebrew could simply be:

`cmake .. -DOPENSSL_ROOT_DIR=$(brew --prefix openssl)`

PR #682, opened 2022-05-11
Add initial support for gnuradio 3.10.

geezer85:
In gnuradio 3.10 the udp_sink "block" moved from gr::blocks to gr::network. This fixes the build failure caused by the move.


Dygear:
Hum. Just to be a little weary of this. The current version of gnuradio on the latest version of Raspberry Pi OS (64bit) Bullseye (Debian 11 based) still downloads `gnuradio arm64 3.8.2.0-14`.

Dygear (approved):
This looks OK to me, as the Pi version (3.8) is handled here.

robotastic:
Awesome! I have to find a way to run GR 3.10 to test against. I will go create a Dev Container to test inside of.

Dygear:
Weirdly even Debian (both the arm64 and amd64 builds) still has 3.8 as their default installed version. I just installed Debian on a MiniForum AMD 5800H and it has the exact same version of the Raspberry Pi 4s that I have setup.

Dygear:
I just decommissioned 4 Raspberry Pi 4s. I'm going to install Ubuntu 22.04 LTS (arm64) onto one of them and try to do the build. I'll report back once done.

Dygear:
Looks like the iomanip include in openmhz and rdio are not actually needed as well. But it builds cleanly with and without them.

plugins/openmhz_uploader/openmhz_uploader.cc
```diff
@@ -1,5 +1,8 @@
 #include <curl/curl.h>
 #include <time.h>
+#if GNURADIO_VERSION >= 0x030a00
+#include <iomanip>
```
Dygear, line 4: Does not seem to be needed, but also doesn't hurt.

geezer85:
I included it because on my setup I got a build error " 'setprecision' is not a member of 'std' " and including iomanip solved it. I only encountered it with GR 3.10. I'm building on Arch Linux ARM using the generic aarch64 image. 

Another header must be including it on the Pi, but not in Arch. It's probably more correct to check if the header was already included rather than check the GR version though since it's not really a GR issue.

robotastic:
Oh interesting... what is a good way to solve this? We could just add iomanip.h with a check and it will just be ignored if it is being included through other paths.
https://stackoverflow.com/questions/12453557/setprecision-is-not-a-member-of-std

geezer85:
The iomanip header should have include guards to prevent multiple inclusions.

robotastic:
Awesome!! Thanks for making these changes

PR #684, opened 2022-05-14
Remove extra characters

dechilders:
These snuck in from commit 6d0a0f1fbce30a7c0b523d2009e121312210e93f -- referenced in https://github.com/robotastic/trunk-recorder/issues/680#issuecomment-1124296176


robotastic:
That is so weird! I wonder how I managed to commit that... let alone type it. Thanks for the fix

Issue #683, opened 2022-05-12
Should Raspberry Pi Users Change the Buster Name to Bullseye?


Dewey3:
I was just pursuing over the Raspberry Pi setup instructions when I realized that a part of the setup reads to add  `deb http://www.deb-multimedia.org/ buster main non-free`  to /etc/apt/sources.list.  Since the Raspberry Pi has been using Bullseye for a while now, that made me wonder if I should change `buster` to `bullseye`, so I visited the http://www.deb-multimedia.org/ webpage.  According to the webpage, screenshot attached, we should now be using `bullseye` instead of `buster`.  Should we now be using `bullseye` instead of `buster` on the line that we add to /etc/apt/sources.list?

![20220512-161730_Image](https://user-images.githubusercontent.com/36827358/168162076-b9e4e30b-494f-4676-b3da-770066c57b9b.jpg)

Dygear:
Ha! I'm actually making a PR for that right now. Along with general house keeping stuff for the Buster -> Bullseye transition and the newer software stack (including using 64bit by default).

PR #685, opened 2022-05-15
Raspberry Pi Bullseye & Ubuntu 22.04 Updates.

Dygear:
Updates the Pi documentation to point to the Bullseye 64 bit build and also the Ubuntu 22.04 updated versions. Resolves https://github.com/robotastic/trunk-recorder/issues/683.

### DIFF
--- a/docs/INSTALL-PI.md
+++ b/docs/INSTALL-PI.md
@@ -77,7 +77,8 @@ sudo apt -y install libssl-dev openssl curl git fdkaac sox libcurl3-gnutls libcu
 Configure RTL-SDRs to load correctly:
 
 ```bash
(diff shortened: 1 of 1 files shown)


robotastic:
Oh Awesome! I will have to give the 64-bit version a try and see if there are perf improvements. Thanks

Dygear:
I think that there is a pref increase, but I'll find out for sure once I have a stable platform across all Pi 4s that I have in production. I have 4x 128GB SD Cards on the way, and 4x FLIRC cases on the way to make sure all units are running the exact same hardware. I've been using a mix of SD Cards (32GB, 64GB) and I'm going to standardize on 128GB cards. I also have 2 Non-FLIRC cases that I'm replacing. After that, I can do a better Apples to Apples (or Pies to Pies) comparison.

PR #695, opened 2022-05-30
Added debian notes to PI install.

Dygear:
Debian works too!

### DIFF
--- a/docs/INSTALL-PI.md
+++ b/docs/INSTALL-PI.md
@@ -101,7 +101,7 @@ sudo apt upgrade
 ```
 
 ```bash
-sudo apt install -y apt-transport-https build-essential ca-certificates fdkaac git gnupg gnuradio gnuradio-dev gr-osmosdr libboost-all-dev libcurl4-openssl-dev libgmp-dev libhackrf-dev liborc-0.4-dev libpthread-stubs0-dev libssl-dev libuhd-dev libusb-dev pkg-config software-properties-common cmake sox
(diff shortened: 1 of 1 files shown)


robotastic:
Should the Debian instructions just be added to the Linux page? Is there anything different between an Ubuntu vs Debian install? I am not sure if folks would check this page if they are not using a Pi

Dewey3:
Not an answer to the question, but Raspberry Pi OS is a custom Debian tailored to the Raspberry Pi (they're both Bulleye), so I'm not sure there are any differences concerning Trunk-Recorder.

Dygear:
> Not an answer to the question, but Raspberry Pi OS is a custom Debian tailored to the Raspberry Pi (they're both Bulleye), so I'm not sure there are any differences concerning Trunk-Recorder.

Yeah, fair, but I thought that maybe the Raspberry Pi people held some packages back because of the maturity of the platform or something. I thought that maybe with the amd64 was more mainstream that it might have more up to date packages.

Dygear:
> Should the Debian instructions just be added to the Linux page? Is there anything different between an Ubuntu vs Debian install? I am not sure if folks would check this page if they are not using a Pi

The version of GNU Radio that Debian and Raspberry Pi install are the same. Ubuntu has a newer version so I'm slightly worried that the steps would be different. As Debian is the common source for Ubuntu, Pop_OS, and Raspberry Pi OS I would at some point hope to see them merged if it doesn't cause to much trouble for the writer and reader.

Dygear:
Added `libsndfile1-dev` to Ubuntu 22.04 installed packages to allow cmake to run to completion.

PR #696, opened 2022-06-05
Update INSTALL-LINUX.md

joe00dev:
Changed libuhd3.15.0 to libuhd4.1.0 for Ubuntu 22.04.


robotastic:
oh interesting! I had tried compiling in a docker container, but it looks like it just uses `libuhd-dev` which may work, but better to refer to the exact library.

PR #698, opened 2022-06-05
Include boost/filesystem header

geezer85:
Without this include I get an error that boost::filesystem::create_directories is not declared while compiling this file.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -191,6 +191,9 @@ if(NOT Boost_FOUND)
 endif()
 
 ADD_DEFINITIONS(-DBOOST_ALL_DYN_LINK)
+if(NOT EXISTS /usr/include/spdlog/fmt/bundled)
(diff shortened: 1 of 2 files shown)


robotastic:
That is weird - is this with Arch Linux?
If so, I can try building it using this container: https://github.com/robotastic/trunk-recorder/blob/master/utils/Dockerfile.arch-latest.dev

geezer85:
This is with Arch Linux. I see the error on an ARM SBC and in that container. I also ran across an issue with Arch's spdlog not bundling fmt. The suggested solution from a quick search was to define SPDLOG_FMT_EXTERNAL in cmake. I wrapped it in a test for the directory the header would be in so as to not break anyone else. For reference see [this old issue](https://github.com/gabime/spdlog/issues/1146).

robotastic:
Cool! thanks for tracking this down

PR #700, opened 2022-06-08
Improved Talkgroup Priority Logging

tadscottsmith:
This adds the ability to assign a talkgroup a priority of -1 to ensure that it is never recorded. This also cleans up the log messages for talkgroup priority based actions to differentiate between when a talkgroup is purposely not recorded (informational) and a talkgroup is not recorded due to a lack or recorders (error).


Dewey3:
Thank you.  Can I continue using 255, or do recommend the -1 from now on?

taclane:
> Thank you. Can I continue using 255, or do recommend the -1 from now on?

Yes, and functionally they'll still both do the same thing.  The advantage of -1 is that it's easier to understand than picking an arbitrarily-large number, and it helps trunk recorder better log why I call isn't being recorded.  For example, if:

TG priority = -1
`Not recording talkgroup. Priority is -1.`

TG priority > # free recorders
`Not recording talkgroup. Priority is 255 but only 6 recorders are available.`

Dewey3:
Ahh, thank you both... I like!

tadscottsmith:
@robotastic I know you're a busy guy, but I'm just curious if there's something that concerns you with the change of behavior here, or if there's anything you want me to change to get it committed? I think this would be a great help for those of us who are really only monitoring error level messages.

robotastic:
Sorry!! This slipped by me. This looks like a good change to me. I had been recommending just to use a high number, but you are correct that it is better to differentiate between the 2 different cases. thanks for adding this!

PR #717, opened 2022-08-13
Add a trailing / to the cp command

DeveloperNeon:
I accidently just overwrote my /etc/systemd/service folder by doing sudo cp trunk-recorder.service /etc/systemd/service without a trailing /


Dygear:
Yikes

Issue #403, opened 2020-09-26
Multiple Airspy Support Confirmation


blantonl:
Can anyone confirm that multiple Airspy support has been patched into the latest version of Trunk Recorder?

YagiBear:
I am looking for the same thing. I used docker to load my install and I am not sure how I could rebuild it with the older gnu radio library required for multiple airspy devices

gonzoblue:
would also like to know as well before I buy a second one...

leee:
trunk-recorder doesn't include gr-osmosdr in our own tree, and users install it from their usual package archive sections (e.g. debian main or ubuntu universe). Packages built by volunteer maintainers/developers for those "official/community" archives tend to stay as close as they can to upstream projects, with patches for platform build/install compatibility and not functional changes.

Likewise, trunk-recorder is downstream of several libraries and projects, including gr-osmosdr and op25. op25 is included in tree because a lot of changes were made to interface with trunk-recorder, that would never make it back upstream. gr-osmosdr hasn't been touched, and (relatively) small changes like this one likely aren't worth adding and forking off our own tree for.

There was some work and a patch suggested for multiple airspy support [in this upstream bug](https://osmocom.org/issues/5144).

If you're comfortable patching and building your own source, you can use the patchfile from the mailing list email linked in that bug (and hope it still applies cleanly and works for gr3.8 - maybe someone can confirm, or manually make the changes yourself). Or you could ask the maintainers of the package across the various platform archives to apply the patch and make available downstream, not as likely.

The easiest request would be to ask for the change to be taken in at the project.

When you use a patched gr-osmosdr (regardless of how it's made available, either by yourself, from a downstream package archive, or from the project itself upstream), trunk-recorder is agnostic to gr-osmosdr settings, and so configuring additional radios as specified by the patch will work out of the box.

If you wish to see this feature made available, asking for the feature in that upstream bug is the best course of action, and I think this issue can be closed. (@robotastic)

If someone tests and confirms that the patch works, then a temporary install note about osmosdr multiple airspy devices with instructions could be added to the wiki until the upstream project (hopefully, eventually) accepts that patch or similar.

blantonl:
I've successfully patched gr-osmosdr with this patch and then compiled trunk recorder

https://lists.osmocom.org/pipermail/osmocom-sdr/2016-April/001446.html

Works great with trunk recorder and is part of the Broadcastify Raspberry Pi Image which includes this compiled version.

[...]

PR #727, opened 2022-09-14
Build gr-osmosdr in Docker with multiple Airspy source patch

EricTendian:
This PR updates how gr-osmosdr is installed in the Docker image for trunk-recorder, to compile gr-osmosdr with a patch mentioned in https://github.com/robotastic/trunk-recorder/issues/403 for allowing multiple Airspys to be selected via serial number. By having this as part of the Docker image, it makes it a lot easier to use multiple Airspy devices without needing to follow a complicated process. As a bonus, gr-osmosdr will now support all the different SDRs it's capable of using (including the newish XTRX), since some extra dev packages were installed that are used in the build process.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,28 +8,52 @@ RUN apt-get update && \
     apt-transport-https \
     build-essential \
     ca-certificates \
+    cmake \
+    docker.io \
     fdkaac \
     git \
-    docker.io \
     gnupg \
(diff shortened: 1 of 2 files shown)


robotastic:
Interesting! I didn't know about this patch and folks have been asking. Let me go pull this in and cut a new version.

Issue #691, opened 2022-05-18
Trunk-Recorder on Ubuntu 22.04


dgwood12:
Is there support for Ubuntu 22.04 (64-bit)? I tried installing trunk-recorder on a new install of Ubuntu 22.04 and encountered some errors when running the following command:
cmake ../trunk-recorder

The here's what the command did:

-- Build type not specified: defaulting to release.
-- Using GMP.
-- Found Boost: /usr/lib/x86_64-linux-gnu/cmake/Boost-1.74.0/BoostConfig.cmake (found suitable version "1.74.0", minimum required is "1.74.0") found components: date_time program_options system regex thread unit_test_framework 
-- User set python executable /usr/bin/python3
-- Extracting version information from git describe...
-- Using GMP.
-- Checking for module 'sndfile'
--   No package 'sndfile' found
-- Could NOT find SNDFILE (missing: SNDFILE_LIBRARIES SNDFILE_INCLUDE_DIRS) 
-- GnuRadio Version: 200705
-- Found gnuradio-uhd: /usr/include, /usr/lib/x86_64-linux-gnu/libgnuradio-uhd.so
-- Checking for module 'gnuradio-osmosdr'
--   Found gnuradio-osmosdr, version 0.2.0.0
-- gnuradio-osmosdr not found.
-- Found gnuradio-osmosdr: , gnuradio-osmosdr;gnuradio-blocks;gnuradio-runtime;gnuradio-pmt
-- Could NOT find GnuradioOsmosdr (missing: GNURADIO_OSMOSDR_INCLUDE_DIRS) 
-- Configuring Boost C++ Libraries...
-- Found Boost: /usr/lib/x86_64-linux-gnu/cmake/Boost-1.74.0/BoostConfig.cmake (found version "1.74.0") found components: thread system program_options filesystem log_setup log regex random 
-- AMBE Tone Synthesis Enabled
-- All libraries:
-- Configuring done
CMake Error at CMakeLists.txt:338 (add_executable):
  Target "trunk-recorder" links to target "sndfile::sndfile" but the target
  was not found.  Perhaps a find_package() call is missing for an IMPORTED
  target, or an ALIAS target is missing?

CMake Error at lib/op25_repeater/lib/CMakeLists.txt:110 (add_library):
  Target "gnuradio-op25_repeater" links to target "sndfile::sndfile" but the
  target was not found.  Perhaps a find_package() call is missing for an
  IMPORTED target, or an ALIAS target is missing?

[...]

PR #733, opened 2022-11-01
Fixes issue mentioned in #691

Dygear:
Adds `libsndfile1-dev` as soft dependency for ubuntu / debian systems. (In my case Pop_OS.)


robotastic:
Thanks!

PR #735, opened 2022-11-21
Fix logging when all transmissions are removed.

tadscottsmith:


robotastic:
Good catch! That is a corner case I forgot about.

PR #738, opened 2022-11-29
Multiple CMake fixes and version updates to resolve compiler warnings

galenguyer:
The output of cmake and make was pretty noisy so I updated some stuff to clean that up. Change list:

* Boost complains about the use of global placeholders. Defining `BOOST_BIND_GLOBAL_PLACEHOLDERS` retains the current behavior
* The declaration of `SPDLOG_FMT_EXTERNAL=ON` conflicts with the declaration of `SPDLOG_FMT_EXTERNAL` in the spdlog CMake instructions. Making the declaration identical resolves the conflict (removing the declaration means it doesn't get built right).
* Compiling on Arch with Boost 1.80 produces deprecation warnings for C++ standards below 14 that will take effect in July 2023. C++17 is the newest supported C++ standard and is supported by Debian 10 and newer as well as Ubuntu 20.04 and newer. 
* Updating C++ to C++14 or newer leads the version of Eigen vendored to throw deprecation warnings for `std::unary_negate` and `std::binary_negate`, which are deprecated in C++17 and removed in C++20. The vendored version of Eigen is at least 5 years old, and the deprecation warning is fixed in Eigen 3.3. Eigen 3.4 has also been released and appears to work without issue, so I've updated the vendored library to that version.
* The OsmoSDR include directory is never found because the wrong pkg-config name is used and the wrong folders are searched in. This is resolved.


robotastic:
I am liking where this is going! Thanks!

robotastic:
Amazing - let's do this. I made a few small tweaks to also cleanup some of the code related warnings.

gisforgirard:
i have nothing to add i just wanted to say wow what an impressive pull request looks like quite a bit of work went into this thanks Galen!

PR #739, opened 2022-12-03
Set d_call to null to prevent potential use after free

galenguyer:
plugman_audio_callback in p25_recorder_decode is sometimes call when d_call is an invalid pointer. This seems to be some form of race condition where d_call has been freed immediately before the audio callback is called, causing a segmentation fault in the simplestream plugin on line 63 (`call->get_system()`). By setting d_call to null when `p25_recorder_decode::stop()` is called,  we can check for it being invalid before a plugin ever sees it, preventing a use-after-free segmentation fault.


robotastic:
Good catch! For the plugin events at the end of a call, they get a copy of the Call info because the call object will get deleted. We could look at doing something similar, but it is probably easier to just check for Null. Especially since this hook gets called so often.

PR #741, opened 2022-12-04
Documenting maxDuration config param

EricTendian:
I noticed that there was a `maxDuration` configuration parameter in the code, but it wasn't actually documented unlike the `minDuration`. This adds a row to the configuration table to document it.


robotastic:
good catch!

Issue #726, opened 2022-09-07
decode metadata in conventionalDMR mode


ZeroChaos-:
Trunk recorder should track the following metadata for conventional DMR:

- [x] caller (already decoded and stored as .srcList[].src
- [ ] talkgroup
- [ ] private/group/all call (all call is a specific value for group call iirc and can be ignored if you prefer)
- [ ] Color Code
- [ ] Slot (all DMR is tdma and should have a slot assigned)

Issue #725, opened 2022-09-07
decode metadata in conventionalP25 mode


ZeroChaos-:
Trunk recorder should track the following metadata for conventional P25:

- [x] caller (already decoded and stored as .srcList[].src
- [ ] talkgroup
- [ ] private/group/all call (all call is a specific value for group call iirc and can be ignored if you prefer)
- [ ] NAC
- [ ] Slot (for tdma p25 phase 2)

chad-d9843:
Was this functionality ever added? It seems I'm unable to capture decoded NACs or packet data frames on P25 conventional systems?

PR #744, opened 2022-12-09
Conventional recorder info for MQTT and status-server plugins

taclane:
Conventional recorders are not displayed in plugins such as MQTT or status-server, or in console logs.

This is addressed by giving each type of conventional recorder (P25c, Analog Conventional, DMR) a separate vector that will be represented in such functions as: `get_recorders()`, `print_recorders()`, `digital_recorder_count()`, and `analog_recorder_count()`.

Analog _conventional_ recorders are currently listed in `analog_recorders` along with _trunked_ recorders. While it seems unlikely that a trunked system would be able to preempt an idle conventional recorder, a separate vector for conventional analog addresses any potential overlap between system types.

To differentiate conventional recorders from their trunked counterparts, `make_xxxx_recorder()` functions accept a `type` string instead of relying on a hard-coded value when the recorder instance is created.  This way, status plugins will not not only display conventional recorders, but can show "P25" or "P25C".

Conventional recorder types are now visible in the periodic console message, along with the center frequency of each source.  This status message has also been indented to match the `Currently Active Calls` display immediately above it.
```
(info)   [ Source 0: 455.562500 MHz ] rtl=01,bias=1
(info)   	[ 9 ] P25C	State: idle
(info)   	[ 10 ] P25C	State: idle
(info)   	[ 11 ] AnalogC	State: idle
(info)   [ Source 1: 770.068750 MHz ] rtl=02
(info)   	[ 0 ] P25	State: available
(info)   	[ 1 ] P25	State: available
(info)   	[ 2 ] P25	State: available
(info)   	[ 3 ] P25	State: available
(info)   	[ 4 ] P25	State: available
(info)   	[ 5 ] P25	State: available
(info)   [ Source 2: 771.506250 MHz ] rtl=03
(info)   	[ 6 ] P25	State: available
(info)   	[ 7 ] P25	State: recording
(info)   	[ 8 ] P25	State: available
```
Any installed MQTT plugins will need to be recompiled after these changes, but no further action is required to display conventional recorders in messages.

### DIFF
--- a/plugins/streamer/streamer_proto_helper.cc
+++ b/plugins/streamer/streamer_proto_helper.cc
@@ -98,7 +98,7 @@ streamer::RecorderInfo* ToRecorderInfo(Recorder *recorder) {
 
     streamer::RecorderInfo* ri;
     ri->set_recorder_num(recorder->get_num());
-    ri->set_recorder_type(recorder->get_type());
+    ri->set_recorder_type(recorder->get_type_string());
     ri->set_source_num(recorder->get_source()->get_num());
     ri->set_id(boost::lexical_cast<std::string>(recorder->get_source()->get_num()) + "_" + boost::lexical_cast<std::string>(recorder->get_num()));
     ri->set_recorder_count(recorder->get_recording_count());
--- a/trunk-recorder/global_structs.h
+++ b/trunk-recorder/global_structs.h
@@ -69,6 +69,15 @@ enum Call_Data_Status { INITIAL,
                         SUCCESS,
                         RETRY,
                         FAILED };
+                  
+enum Recorder_Type { DEBUG,
+                      SIGMF,
+                      ANALOG,
+                      ANALOGC,
+                      P25,
+                      P25C,
+                      DMR };
+
 struct Call_Data_t {
   long talkgroup;
   std::vector<unsigned long> patched_talkgroups;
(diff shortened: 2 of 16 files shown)


robotastic:
I like this approach - I think this could eventually help with #725 and #726 I might try making an edit so ENUM are used with a function to convert them to a string output. That will help prevent errors around mistyping the string.

robotastic:
I made a few edits to add in ENUM - lets give this a try

taclane:
Ah great!  Using an ENUM is probably the better way to go, though it seems to be causing problems downstream with the MQTT and status-server plugins.

status-server seems like it just needs in `recorder.cc`:
```cpp
boost::property_tree::ptree Recorder::get_stats() {
...
  node.put("type", get_type_string());
```
But I can't quite put my finger on where the switch from a type string to enum is causing the MQTT plugins to hang on launch.

robotastic:
Ah - let me go give things a try.... I didn't test as well as I should have

taclane:
Might have just been an issue rebuilding the MQTT plugins.  They seem to be working again after cleaning all the build directories and reinstalling...

The above edit is still needed so that the plugins don't just show an integer in the type field.

```
Client (null) received PUBLISH (d0, q0, r0, m0, 'taclane/feeds/recorder', ... (227 bytes))
taclane/feeds/recorder {
    "recorder": {
        "id": "0_10",
        "type": "5",
        "srcNum": "0",
        "recNum": "10",
        "count": "2",
        "duration": "2.8799999999999999",
        "state": "4"
    },
    "type": "recorder"
```

robotastic:
Awesome - I just checked and it looks like the MQTT plugins are working for me too. I also just cleaned the up a bit, prob worth getting a new copy.

PR #745, opened 2022-12-14
Cleanup dropped samples logging.

tadscottsmith:
When an inactive P25 call is stopped, occasionally there can still be a single tagged sample used to pass the terminate tag that transmission_sink attempts to process, resulting in a "current_call is null" error. This PR would make the log trace-level if only a single sample is dropped, and make it an error-level when more samples are dropped.
```
[2022-12-13 17:36:42.010653] (info)   [ramsey]  923C     - Stopping P25 Recorder Num [1] TG:           EMSMRCC (      2142)   Freq: 852.662500 MHz    TDMA: false     Slot: 0 Hz Error: -926
[2022-12-13 17:36:42.011798] (info)   [ramsey]  923C     TG: 2142     Freq: 852.662500 MHz    - Transmission src: 121148 pos: 0 length: 1.08
[2022-12-13 17:36:42.012359] (error)   [ramsey] 923C     TG: 2142     Freq: 852.662500 MHz    Dropping samples - current_call is null        Rec State: available   Since close: 1
```


robotastic:
I like it! not a huge issue if we don't get all the TDUs.

Issue #723, opened 2022-09-03
P25 Parse error


stevegame:
I was trying to use the trunked-recorder on my laptop with my sdr
i want to use my **rafael tuner**

-----
gr-osmosdr 0.2.0.0 (0.2.0) gnuradio 3.8.1.0
built-in source types: file osmosdr fcd rtl rtl_tcp uhd miri hackrf bladerf rfspace airspy airspyhf soapy redpitaya freesrp 
[INFO] [UHD] linux; GNU C++ version 9.2.1 20200304; Boost_107100; UHD_3.15.0.0-2build5
Found Rafael Micro R820T tuner
-------------->>Using device #0 Realtek RTL2838UHIDIR SN: 00000001
--------------->>Found Rafael Micro R820T tuner
[R82XX] PLL not locked!
[2022-09-03 02:45:55.276783] (info)   SOURCE TYPE OSMOSDR (osmosdr)
[2022-09-03 02:45:55.276860] (info)   Setting sample rate to: 8000000
Invalid sample rate: 8000000 Hz
[2022-09-03 02:45:55.276965] (info)   Actual sample rate: 1024000
[2022-09-03 02:45:55.277010] (info)   Tuning to 8.570000e+08
[2022-09-03 02:45:55.329850] (info)   Gain Stage: LNA supported values: 0 0.9 1.4 2.7 3.7 7.7 8.7 12.5 14.4 15.7 16.6 19.7 20.7 22.9 25.4 28 29.7 32.8 33.8 36.4 37.2 38.6 40.2 42.1 43.4 43.9 44.5 48 49.6 
[2022-09-03 02:45:55.329934] (info)   Max Frequency: 8.609680e+08
[2022-09-03 02:45:55.330008] (info)   Min Frequency: 8.530320e+08
[2022-09-03 02:45:55.356569] (info)   Gain set to: 40.2
[2022-09-03 02:45:55.356686] (info)   Auto gain control is OFF
[2022-09-03 02:45:55.356739] (info)   Setting antenna to [RX]
[2022-09-03 02:45:55.357523] (info)   	 P25 Recorder two-stage decimator - Initial decimated rate: 100000 Second decimated rate: 25000 FA: 6250 FB: 12500 System Rate: 8000000
[2022-09-03 02:45:55.401456] (info)   	 P25 Recorder ARB - Initial Rate: 8000000 Resampled Rate: 25000 Initial Decimation: 80 ARB Rate: 0.96 SPS: 5

---
[...]

PR #753, opened 2022-12-23
adjust Source sample rate to match SDR sample rate

aaknitt:
If the hardware sample rate doesn't match the Source sample rate (from config.json), adjust the Source sample rate to match.


aaknitt:
For #723

robotastic:
This is prob the Best 1 Line PR ever - great idea!

PR #754, opened 2022-12-23
Create log file earlier

aaknitt:
Create log file earlier so that config and source creation info is logged to the file in addition to the console output.

Reference #723


tadscottsmith:
Good idea!

robotastic:
It like it!

rosecitytransit:
This needs to be reopened. The `logging::add_file_log` was duplicated instead of moved, so this effectively doesn't change anything. Also it would be ideal to be able to disable the console output (by not running https://github.com/robotastic/trunk-recorder/blob/7bb5742a88d816d4ce3f8a68053d70e2e9cffbc8/trunk-recorder/main.cc#L1706). There's only a few lines that don't make it into the file log; mostly some that doesn't go through BOOST logging at all.

If you really want, I could open a PR for it.

PR #756, opened 2022-12-28
Enable/Disable config sections

taclane:
For the purpose of troubleshooting or managing a trunk-recorder setup, it is desirable to easily enable/disable portions of the config.json.  The only way to presently do so is to save a backup copy and delete portions of the config, or copy/paste any deleted portions back in.

This change simply adds a `"enabled": false` or `"enabled": true`(default) option to individual `sources`, `systems`, or `plugins`.  Disabled elements of the config file are simply ignored upon startup.


robotastic:
This should really help with debugging - thanks!

PR #757, opened 2022-12-29
broadcastify: add option to disable ssl certificate verification

galenguyer:
Broadcastify let their SSL certificate expire... again...


robotastic:
Sure - why not!

PR #758, opened 2022-12-29
Enable Multi-site mode features.

tadscottsmith:
There are a number of people that have been testing this and hoping to see it in the master branch. I am open to any suggestions if there's something you'd like to see done a different way. This shouldn't impact anyone unless they enable the new configuration options.


robotastic:
This is awesome! I am going to have to spend some time really looking through it - but it sounds like a lot of folks have gotten it to work well.

tadscottsmith:
Great! Let me know if you have any suggestions or want me to update anything. This version primarily helps conserve recorder resources by only recording the call from one site. I also think it would be worthwhile to have an option to record the same call from both sites, and then use some logic in the call concluder to take the best transmissions from either site to attempt to create the best quality call. 

One big thing I'd like your input on is tracking the reason that calls are in a monitoring state. I initially implemented a separate MonitoringState struct (as linked below), but I'm curious if you'd prefer to have everything in the original "State" or if you're onboard with breaking out the additional monitoring states?  A few future use cases I could see are starting a monitored call if a recorder becomes available mid call, or creating metadata files / plugin calls for certain types of monitored calls that someone doesn't chose to record. https://github.com/tadscottsmith/trunk-recorder/blob/88b4289d0713d796411eeb1ccdb9e5f0d96600a2/trunk-recorder/state.h#L13-L20

tadscottsmith:
Let me know if it would be helpful to break this up into multiple PRs. I could do one to add monitoring state, and a separate one to add the optional multi-site config.

robotastic:
I have made it half way through, and so far it is making sense! I just got distracted working on the frontend stuff.... let me see if I can finish it up tonight

robotastic:
This looks great! I think Monitoring state should help out things in general. And it should work pretty much out of the box for most folks, and you can make it work better with a little configuration

PR #759, opened 2022-12-29
Update signaling decoders to print logs via boost info log instead of plain stdout

EricTendian:
I noticed that the signaling decoders were printing out decoded signals to stdout, instead of via a `BOOST_LOG_TRIVIAL` call. This can get pretty noisy if there's a lot of transmissions being decoded. By using the info log, someone can set a different loglevel to exclude them.

This changes the logs from:

```
{"type":"MDC1200","timestamp":"1672355479","op":"01","arg":"00","unitID":"5bad","ex0":"00","ex1":"00","ex2":"00","ex3":"00"}
```

to:

```
[2022-12-29 23:11:19.361464] (info)   {"type":"MDC1200","timestamp":"1672355479","op":"01","arg":"00","unitID":"5bad","ex0":"00","ex1":"00","ex2":"00","ex3":"00"}
```

Besides this change, I also added `curl` to the Dockerfile since it can be useful for upload scripts that need to call APIs (and there's no other http client currently present).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,6 +9,7 @@ RUN apt-get update && \
     build-essential \
     ca-certificates \
     cmake \
+    curl \
     docker.io \
     fdkaac \
     git \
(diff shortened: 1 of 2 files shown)


robotastic:
Good catch! It also looks like this should address #762

PR #764, opened 2023-01-21
Adding support for regex in unitTagsFile

EricTendian:
For large radio systems, it can be impractical to list out every possible permutation of a radio ID in the `unitTagsFile`. Often radio systems have IDs that follow particular patterns which can be modeled in a regular expression.

This PR adds support to the `unitTagsFile` for regular expressions, such that all units matching the same regex can be tagged with the corresponding name. Capture groups are supported in the name as well, so a name of "Engine $1" in the file can get translated to "Engine 225".

I've updated the configuration docs with more information on how to use this.

### DIFF
--- a/docs/CONFIGURE.md
+++ b/docs/CONFIGURE.md
@@ -26,31 +26,31 @@ The amount of tuning error is -14500Hz, so that would go under **error:** for th
 
 **NOTE:** In some instances, an alternative is to use `ppm` correction rather than the `error` configuration option.
 
-Alternatively, you can use this tool here: http://garvas.org/trunk-recorder/ to obtain RTL-SDR dongle/array configurations. 
+Alternatively, you can use this tool here: http://garvas.org/trunk-recorder/ to obtain RTL-SDR dongle/array configurations.
 
 ### Gain
 
-After you have figured out the amount of tuning error, you now need to find the optimal amount of receiver gain to use. 
+After you have figured out the amount of tuning error, you now need to find the optimal amount of receiver gain to use.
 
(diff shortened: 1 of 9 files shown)


docs/CONFIGURE.md
```diff
@@ -26,31 +26,31 @@ The amount of tuning error is -14500Hz, so that would go under **error:** for th
 
 **NOTE:** In some instances, an alternative is to use `ppm` correction rather than the `error` configuration option.
 
-Alternatively, you can use this tool here: http://garvas.org/trunk-recorder/ to obtain RTL-SDR dongle/array configurations.
```
EricTendian, line 29: Ignore most of these changes, my IDE is just getting rid of trailing whitespace.

docs/CONFIGURE.md
```diff
@@ -401,3 +401,18 @@ The **Enable** Column is optional and defaults to *True*. It only needs to be ad
 | --------- | --------- | -------- | ------------- | ---------------------- | ------ | ------ | ------------------- |
 | 300       | 462275000 | 94.8 PL  | Town A Police | Town A Police Dispatch | Police | Town A |                     |
 | 325       | 462275000 | 151.4 PL | Town B DPW    | Town B Trash Dispatch  | DPW    | Town B | False               |
+
+
+## unitTagsFile
```
EricTendian, line 406: Here's the actual doc update.

taclane:
Very nice!  I tried something like this a while back with just a simple wildcard character like: `1427???, State Police` and it was really cool to start seeing meaningful unit IDs come across in Rdio-Scanner. This is a way better solution.

robotastic:
This is really slick!! Nice work @EricTendian

PR #768, opened 2023-01-30
Use more compatible cmake command

dotsam:
`add_compile_definitions` was added in cmake 3.12.4, but is just some syntactical sugar for `add_definitions`. Fall back to that command to keep compatibility with the defined minimum cmake version (3.2). This also maintains compatibility with Ubuntu 18.04, as it ships with cmake 3.10.2.


robotastic:
CMAKE is a dark art - thanks for fixing this!

PR #772, opened 2023-02-09
fix for elusive crashing issue

gisforgirard:
see https://www.ruby-forum.com/t/s-d-bufsize-failed/59342/4

randomly this issue started raising its head again...

and on further investigation found this similar issue that involved almost the exact same pseudo-code,

I made the change and it appears to have stabilized a crashing issue that has been persisting for the past couple of days out of nowhere for some reason, appears from the debug logs I posted before to be the same issue, will delete this if it doesn't solve it upon further testing


gisforgirard:
i lied it messed up the audio

PR #776, opened 2023-02-13
Update Homebrew instructions

taclane:
Update the MacOS Homebrew instructions to be current as of Feb 2023, noting the packages available under MacOS Ventura.

Install instructions under Homebrew better mirror those used for MacPorts, and note that the location of `openssl@3` differs by macOS system type:

>This script installs Homebrew to its default, supported, best prefix (/usr/local for macOS Intel, /opt/homebrew for Apple Silicon ...)
>-https://docs.brew.sh/Installation

The "Homebrew way" of using `-DOPENSSL_ROOT_DIR=$(brew --prefix openssl@3)` is proposed to make things simple and cover both architectures.


robotastic:
That is a great trick for locating the openssl libs!!! thanks for cleaning this up!

PR #777, opened 2023-02-15
dropping more samples than it should

gisforgirard:
shouldn't this only be dropping samples if the state is stopped and record_more_transmissions is false? it seems like that would follow the STATES.md more closely but I don't know what I'm doing so there's a good chance I don't know what I'm talking about


robotastic:
I am not sure if this would change anything - The recorder should only enter the STOPPED state after record_more_transmissions is set to `false`. But who knows, maybe something is messed. Have you tried adding a print statement to see if record_more_transmissions is ever `true`?

gisforgirard:
> I am not sure if this would change anything - The recorder should only enter the STOPPED state after record_more_transmissions is set to `false`. But who knows, maybe something is messed. Have you tried adding a print statement to see if record_more_transmissions is ever `true`?

no, I hadn't thought of that, but now that I have it looks like you are probably right, I need to stop bothering y'all wasting your time thinking I know what I'm doing because I clearly don't

PR #784, opened 2023-03-03
Console log cleanup

taclane:
A few minor fixes for console logs to aid in consistency and readability.

- Correct several lines that displayed blue due to unterminated ANSI color codes
- The timing of a lone `Setting TDMA to: true/false` lacks context at the `info` level, and the information is displayed more effectively along with system and talkgroup info in:
    `Starting P25 Recorder Num [N]	TDMA:true/false`
- Drop a couple instances of "_impl" that is inconsistent with other info-level logging


robotastic:
Looks good - thanks for helping tidy things up!

PR #785, opened 2023-03-03
UnitID alpha improvements and decode error display

taclane:
This commit will display UnitID alpha in console messages to improve the view of trunk-recorder operations at the `info` logging level and aid in any UID regex debugging.  UnitID alpha was previously displayed in trunk-recorder 3.x console logs, and this commit would restore some of that functionality.

When a call ends, Call_Concluder will now display known UnitID alpha and any reported errors/spikes for that transmission.  Call times are formatted to two decimal places to avoid the initial `pos:  0` creating text alignment issues.

If `hideEncrypted` is false (default), UnitID and alpha will also be displayed to provide some insight into the nature of those transmissions.

Per discussion re: #764 with @EricTendian, the `format_all` flag is included to allow complex regex matches and conditional substitutions with the `unitTagsFile` in addition to the current behavior.

### DIFF
--- a/trunk-recorder/call_concluder/call_concluder.cc
+++ b/trunk-recorder/call_concluder/call_concluder.cc
@@ -276,7 +276,20 @@ Call_Data_t Call_Concluder::create_call_data(Call *call, System *sys, Config con
       continue;
     }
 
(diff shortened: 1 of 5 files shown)


robotastic:
Great changes! sorry it took me so long to merge in

Issue #780, opened 2023-02-24
multiple failures with CXXFLAGS="-D_GLIBCXX_ASSERTIONS"


ZeroChaos-:
-D_GLIBCXX_ASSERTIONS is a default CXXFLAG on multiple linux distros, it performs runtime bounds checking in C++
https://wiki.gentoo.org/wiki/GCC_optimization#Overflow_protection

When this is enabled, quite a few bugs present immediately (https://github.com/robotastic/trunk-recorder/issues/779).  I would recommend (at a minimum) adding this flag for the debug builds so it's used during testing and these bugs can be identified.  A min test case is here:

```
CXXFLAGS="-D_GLIBCXX_ASSERTIONS" cmake ..
make VERBOSE=1
```

I'm happy to make  a PR adding the flag to CMakeLists.txt if desired, but I wanted to open this discussion as to why and confirm it would be desired.

ZeroChaos-:
https://github.com/robotastic/trunk-recorder/issues/783

ZeroChaos-:
I think these are all fixed now

PR #793, opened 2023-04-04
add -D_GLIBCXX_ASSERTIONS to CXXFLAGS

ZeroChaos-:
Many linux distributions have switched this on by default, but TR does not have it on by default and so bugs are slipping through and hitting users.  Turn on bounds checking by default so that bugs are found quicker.

This WILL break users.  I'm sure of it.  It will cause program exit when running broken code, example being https://github.com/robotastic/trunk-recorder/issues/783 The performance penalty of this is minimal (none on modern distros where this is already teh default) and the bounds checking causing failure is better than simply running objectively wrong code.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -223,7 +223,7 @@ set(CMAKE_CXX_FLAGS_DEBUG "-Wall -Wno-unused-local-typedef -Wno-deprecated-decla
 
 SET(CMAKE_CXX_STANDARD 17)
 
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -pthread -Wno-narrowing -fvisibility=hidden -fPIC")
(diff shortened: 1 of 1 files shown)


ZeroChaos-:
https://github.com/robotastic/trunk-recorder/issues/783
https://github.com/robotastic/trunk-recorder/issues/780

robotastic:
This is 🔥! I like it, but I probably have to get a host of people to run this on test installs first before turning it on for everyone. I am scared of all the ways it will break. 

How about I create a branch with this on, and we can collect all of the fixes in that branch. Once we are getting good uptime with a number of different system types, we can mainline it?

ZeroChaos-:
I leave that up to you.  I will reiterate now that this will only cause breakage on objectively wrong code.  It may be better to find it all quickly and rip off the band aid.  Additionally, this is a default use flag for most modern major distros, so time is short here and more and more users will be hitting runtime errors.  I support your solution, but I do feel pulling into master will get much wider testing much faster.  You may choose to fix the bug I already found first ;-)

robotastic:
This worked great! merged it in, but commented it out for the Prod branch

PR #794, opened 2023-04-07
OpenMHz Uploader update

robotastic:
Changes the way Error Count and Spike Count are sent to OpenMHz. It removes the Freq List and just sends the values as part of the main form. There is only one freq per call now, so no need for a list anymore.


rosecitytransit:
This should have gone away once the retune function was eliminated. I know it could break other consumers of the data, but you could also eliminate putting it in the json files in call_concluder.cc. The error and spike counts could go in the transmission entry. Or if you just want to log them per call, you could not reset them every transmission (in p25_frame_assembler_impl.cc / p25p1_fdma::reset_rx_status())

PR #796, opened 2023-04-11
Ignore missing files

devicenull:
Occasionally, trunk-recorder tries to combine two wav files into one, but one of them doesn't actually exist. This fixes the issue by checking for file existence before sending files to sox - which will preserve any audio that exists in the remaining file.

This situation was causing a `Failed to combine recordings, see above error. Make sure you have sox and fdkaac installed.` error occasionally, and resulted in the audio not being properly saved.

### DIFF
--- a/trunk-recorder/call_concluder/call_concluder.cc
+++ b/trunk-recorder/call_concluder/call_concluder.cc
@@ -156,13 +156,21 @@ Call_Data_t upload_call_worker(Call_Data_t call_info) {
     std::string shell_command_string;
     std::string files;
 
+    struct stat statbuf;
     // loop through the transmission list, pull in things to fill in totals for call_info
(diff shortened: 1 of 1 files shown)


robotastic:
Interesting! I wonder what is causing the missing files... let me know if you have any clues. Maybe there is some delay in writing.

Either way, this is a good error check. Thanks!

taclane:
Not sure if if's just me, but it looks like `call_concluder.cc` might need a:
```
#include <sys/stat.h>
```
to get this to compile on Ubuntu and pass the github Buildx check.

robotastic:
Should be fixed, thanks to @taclane

robotastic:
Or more accurately - a similar, but different bug was also addressed.

PR #797, opened 2023-04-18
Adding support for SpyServer to Dockerfile

EricTendian:
This PR makes a few improvements to the Dockerfile by adding support for SpyServer, via a fork of gr-osmosdr. I just submitted a PR to it which got it back in sync with the upstream repo, as well as added the AirSpy serial number patch, so we no longer need to do that as part of the build process (which is why I'm removing the patch).

This should allow the Docker setup to have support for even more SDRs.


EricTendian:
(need to fix CI, apparently)

robotastic:
Ugh! This looks like some weird build bug for gr-osmosdr.

robotastic:
ACtually - it may have nothing to do with a change you made and was a separate that was fixed. Could you make some sort of trivial change that would retrigger the build?

EricTendian:
@robotastic this was an issue with my changes to the gr-osmosdr fork, I submitted a PR to that repo which will fix the issue, and then I'll be able to open this PR up again.

EricTendian:
@robotastic this is finally ready for a review now that the gr-osmosdr fork has been updated. I ended up taking out the SDRPlay changes since that's technically already available in gr-osmosdr.

PR #798, opened 2023-04-21
Spelling fixes and additional call stat reporting

taclane:
A few minor fixes:

`call->get_stats()` should provide downstream plugins (such as MQTT) a call's number as a shorter alternative to the long `id`.  

Calls stats should also provide the `monitoringState` to better identify outlier trunk recorder events such as duplicate calls or the lack of available recorders/sources.  This would give users more detailed state information than `state` and `recState` currently provide.

Removed a leftover "Call_impl" reference to be consistent with other `info` level console text.

Corrected the spelling of the "elapsed" key used in call stats.


robotastic:
Great upgrades!! Thanks

PR #799, opened 2023-04-25
Standardize Talkgroup Class

tadscottsmith:
This would standardize the talkgroup class to user getters/setters much like many of the other classes. This would also allow access to modify many of the talkgroup's settings from either within trunk-recorder or from the plugins.

I've modified the internal calls and verified the internal plugins work with these additions. This could potentially break any external plugins that are directly accessing the talkgroup members, but it seems like the added functionality would probably outweigh the minor changes needed to update these plugins.


robotastic:
This seems to make sense to me - in theory, it should be backwards compatible because the public variables are still there in case some plugin is still accessing them directly, right?

PR #773, opened 2023-02-10
fix for elusive crashing issue

gisforgirard:
I kinda jumped the gun and posted a pull request the other day without testing it, but I've been running with this change all day and like 50,000 calls, everything appears to be functioning perfectly fine... I think it does seem to be ending calls sooner, but not like any earlier than they should be, just tends to seem like less of a silence gap between calls within a single recording. 

Does anyone know why it would continue processing call audio if terminate_call is true? This seems to have solved the crashing issue I was having, and looking back at a previous issue that I thought was solved, I noticed in the debug info I posted back then, it was crashing while terminate_call was true, so it appears this could have been the actual issue at hand and I didn't notice because for whatever reason this crash happens randomly a bunch of times in a row and then goes away completely, but this time I actually had a chance to track it down and figure it out while it was happening, compile a bunch of versions with extra debug info, and narrow it down to pretty much this exact spot where it was having issues.

Issue #564, opened 2021-11-30
Calls are out of Order


JayManCanada6014:
This is the same thing I posted on Gitter,  there are multiple people in the thread reporting the same.  This has only started say in the last 2-3 days and only seems to impact digital calls. Analog I have not heard the issue on yet.

I'm not sure if this makes sense and if some patch to linux did it or something else. I am getting calls out of order, a lot of them I hear a call, then hear the dispatcher calling for the unit and then the call continues which of course isn't the right order for a call.
I've increased my timeout from 4 to 5 seconds but I was wondering if anyone has any suggestions.

johnrockie:
Yeah can confirm same issues since going from v3 to 4, specially the crosstalk where you hear both FD and PD on a single audio file. I'll try to get examples, the weird thing I have two systems in two different locations, both P25 II only difference is HackRF vs Airspy, HackRF one is the one with issues but I don't believe that's the issue. I provided something on [#535](https://github.com/robotastic/trunk-recorder/issues/535) but will try to get some more.

sally-yachts:
I experienced the same issue when testing an upgrade to 4.x. Here's some relevant messages from Gitter (thread brom 8/13/21:

> So actually I may not be fully converted after all. I'm going to setup another testbed, but i've got users reporting that transmissions within a file are ordered incorrectly - has anyone ever heard of something like this? The only thing that's changed is code upgrade and switching from a LimeSDR Mini to an Airspy.

(Note: I have since tested 4.x using the same LimeSDR Mini and still encountered this issue so I don't believe it to be an issue with the SDR itself)

>From what I'm looking at it looks like the actual problem is that recorders don't seem to be tracking/releasing properly. Calls from different talkgroups are being merged in the same file.

>An example of this can be seen here - https://tp-dev.atlscanner.com/audio/1e935f22-8fd3-41f9-8a6b-77bab69678b9/ . There are calls from multiple talkgroups in this one converstation, including mixing of fire and police calls. What's interesting is that the extra calls don't appear to be written to the metadata - it's like the WAV recorder just keeps getting fed data that should be going to another recorder process.

>To make things even more interesting, that extra audio (like fire audio) destined for another recording process still appears to make it to the proper destination as well, resulting in radio transmissions being duplicated across active recorder processes. Real weird.

>For reference the deployment having the issue is using latest docker image/build as of 8/13/2021, and the version that is working properly is commit 0b1719d from 8/13/2020.

[...]

PR #800, opened 2023-05-02
actual crash fix this time

gisforgirard:
this is the actual crash fix i kept messing up and submitting pull requests for, for real this time

honestly this is probably not the best way to solve it, but there is a case where amt_produce ends up going over 32768 and it crashes the program faithfully every time, and this has appeared to solve it for me without any obvious negative side effects, there is probably a better way to go about doing this and its pretty much a hack, but it's a hack that works at least

### DIFF
--- a/lib/op25_repeater/lib/p25_frame_assembler_impl.cc
+++ b/lib/op25_repeater/lib/p25_frame_assembler_impl.cc
@@ -183,6 +183,10 @@ p25_frame_assembler_impl::general_work (int noutput_items,
         //BOOST_LOG_TRIVIAL(trace) << "P25 Frame Assembler -  output_queue: " << output_queue.size() << " noutput_items: " <<  noutput_items << " ninput_items: " << ninput_items[0];
 
       if (amt_produce > 0) {
+          if (amt_produce >= 32768) {
+            BOOST_LOG_TRIVIAL(error) << "just saved you from a crash, amt_produce = " << amt_produce;
+            amt_produce = 32767; // buffer limit is 32768, see gnuradio/gnuradio-runtime/lib/../include/gnuradio/buffer.h:186
+          }
           long src_id = p1fdma.get_curr_src_id();
           long grp_id = p1fdma.get_curr_grp_id();
           // If a SRC wasn't received on the voice channel since the last check, it will be -1
--- a/lib/op25_repeater/lib/p25_frame_assembler_impl.cc
+++ b/lib/op25_repeater/lib/p25_frame_assembler_impl.cc
@@ -184,7 +184,8 @@ p25_frame_assembler_impl::general_work (int noutput_items,
 
       if (amt_produce > 0) {
           if (amt_produce >= 32768) {
-            BOOST_LOG_TRIVIAL(error) << "just saved you from a crash, amt_produce = " << amt_produce;
+            BOOST_LOG_TRIVIAL(error) << "P25 Frame Assembler -  output_queue size: " << output_queue.size() << " max size: " << output_queue.max_size() << " limiting amt_produce to  32767 ";
+            
             amt_produce = 32767; // buffer limit is 32768, see gnuradio/gnuradio-runtime/lib/../include/gnuradio/buffer.h:186
           }
           long src_id = p1fdma.get_curr_src_id();


Dygear:
When does `amt_produce` go over the 32768 limit? Sounds like there is another bug lurking there somewhere else. This feels a little like a band aide to the problem. Will for example discarding past the 32767th byte cause significant data loss for voice data or something else?

robotastic:
This is definitely interesting! let me see if I can find some suspects. amt_produced should never get that large.

tadscottsmith:
I say it's always a good idea to check for overflows, but I'm pretty sure you've identified the reason you're seeing the issue, and most others aren't: https://github.com/robotastic/trunk-recorder/pull/773#issuecomment-1425881600.

robotastic:
Tracking some research:
**amt_produce** is an `int` which means it can have a range between:  -2147483648 to 2147483647
it is set to the value of the queue holding the output - which is returned as a size_t.

The baffling thing is that amt_produce should never be that big. It is however many samples were processed since the last time things were processed.

Could you uncomment line 183 in that file? I am curious what values you are seeing for the output queue...

robotastic:
oh - you maybe right @tadscottsmith That would explain why so much gets built up... I am honestly surprised this fixes things... and also why that value, it seems like an int could hold a lot more. maybe it is something further down the stream.

faultywarrior:
I suppose it is possible that the queue is getting built up that big due to filesystem writes being held; but it does seem odd.  Usually if you start to out-strip your storage throughput, you just start dropping traffic or get spammed with O's.  FWIW, I capture about a dozen of calls per second on my T-R instance to a NAS which, like @gisforgirard, I have it mounted as a local folder.  I've never seen this crash occur - my TR instance has had months of uptime.  However, my NAS is fully solid-state and has a 10Gb connection to the ESXi host where the T-R instance runs; so its possible that despite the large write volume (my T-R logs are witten to the same NAS); my latency is low enough that I don't build the buffer up big enough to trip this bug.

tadscottsmith:
If my math is right, and 160 samples = 20 ms, then 32768 samples is (32768 / 160 = 204.8 * 20 ms  = 4096 ms)  ~ 4.1 seconds behind.

The max size of an Int_16 is 32767 and there's an awful lot of those in the area that would make me suspicious.

https://github.com/robotastic/trunk-recorder/blob/1718ea5f3eed294de2c5d025858a8fdd33f5188d/lib/op25_repeater/lib/p25_frame_assembler_impl.cc#L181

This line stands out after a quick glance.

https://github.com/robotastic/trunk-recorder/blob/1718ea5f3eed294de2c5d025858a8fdd33f5188d/lib/op25_repeater/lib/p25_frame_assembler_impl.cc#L207

taclane:
> If my math is right, and 160 samples = 20 ms, then 32768 samples is (32768 / 160 = 204.8 * 20 ms = 4096 ms) ~ 4.1 seconds behind.

That sounds about right.  A couple years back [there was an issue](https://github.com/robotastic/trunk-recorder/issues/564) where buffers would back up and it led to call audio falling out of sequence.  60-90k in the queue time-shifted transmissions 8-10 seconds into the next call.

A little curious about "// buffer limit is 32768, see gnuradio/gnuradio-runtime/lib/../include/gnuradio/buffer.h:186".  Which version of gnuradio is this from?

Dygear:
Oh, not flushing to the network is a problem. Didn't think that it would hold onto it in a buffer there, but in a file buffer somewhere else. It could save to the file system (even remote file systems) once the device becomes ready for it.

Sounds like the real fix is making a file save function that holds onto a que of data that it's waiting to write to disk somewhere in the main loop so it doesn't backpressure into this buffer.

robotastic:
I added a few more debug values to the message. I am also checking with a friend to see if he has any ideas... either way, this seems like a good fix, I am merging it in.

gisforgirard:
> I suppose it is possible that the queue is getting built up that big due to filesystem writes being held; but it does seem odd. Usually if you start to out-strip your storage throughput, you just start dropping traffic or get spammed with O's. FWIW, I capture about a dozen of calls per second on my T-R instance to a NAS which, like @gisforgirard, I have it mounted as a local folder. I've never seen this crash occur - my TR instance has had months of uptime. However, my NAS is fully solid-state and has a 10Gb connection to the ESXi host where the T-R instance runs; so its possible that despite the large write volume (my T-R logs are witten to the same NAS); my latency is low enough that I don't build the buffer up big enough to trip this bug.

yeah i have a 10GBASE-T link straight to my NAS but for whatever reason it has weird latency issues where packets won't drop but it will freeze the entire system at the kernel whenever it happens, i have tried different network cards, ethernet cords, kernel settings, and it drives me crazy trying to track down what is going on but still haven't been able to figure it out, it's probably some kind of problem with synology's software but i'm already like $3000 into this stupid thing so maybe i'll go back to truenas for the next build instead...

PR #802, opened 2023-05-04
Fix freq offset variable declarations in trunking messages

taclane:
Frequency offsets calculated by the p25 parser can sometimes be negative, but are currently being stored as unsigned.  While this information does not appear to be used by trunk-recorder, it corrects malformed output when viewing logs at the `trace` level.


robotastic:
Interesting and good catch!!

PR #809, opened 2023-05-08
instanceId and fixes for active call handling.

TheGreatCodeholio:
Added a global config named instanceId that is a string. It is currently sent with recorder and call data for stats so you can decipher which trunk-recorder instance it came from. 

Added a new vector to hold only active calls. When a call is started it is added to it, and when it ends it is removed from it. This is so when announced via plugman_active_calls you get only actual active calls on conventional. Previously conventional was adding all active recorders as an active call, which was all of them due to the nature of how conventional is handled. 

Moved active call update for plugman. Put in the loop with call rate. Previously was not triggering for conventional calls, due to them being handled in a separate function.

### DIFF
--- a/trunk-recorder/call_impl.cc
+++ b/trunk-recorder/call_impl.cc
@@ -13,7 +13,7 @@ std::string Call_impl::get_capture_dir() {
   return this->config.capture_dir;
 }
 
-int Call_impl::get_instance_id() {
+std::string Call_impl::get_instance_id() {
     return this->config.instance_id;
 }
 /*
--- a/trunk-recorder/call_impl.cc
+++ b/trunk-recorder/call_impl.cc
@@ -12,6 +12,10 @@
 std::string Call_impl::get_capture_dir() {
   return this->config.capture_dir;
 }
+
+int Call_impl::get_instance_id() {
+    return this->config.instance_id;
+}
 /*
 Call * Call::make(long t, double f, System *s, Config c) {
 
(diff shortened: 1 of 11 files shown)


taclane:
As an alternative light-weight option, a lot of this could be handled by a plugin without adding overhead to the call flow.  With the current plugin API it is possible to:
- Get the `instance_id` (or any other key) from the config.json and append it to all outgoing MQTT messages.
- On `calls_active()`, pare down the calls vector to remove conventional calls with zero length (active idlers that have not yet started recording) before sending the information along.

But I do entirely agree that the irregular reporting of conventional calls would be helped by a periodic `plugman_active_calls(calls)`.  Sending it at the same frequency as the control channel rates is logical, and it's much easier for plugins to work with.

robotastic:
Yep - I like the general intent on this - both seem like good ideas.

Would it be possible to instead set the MQTT Client ID for each Trunk Recorder system to help tell them apart?

I would be better to not have a second vector trying to keep track of the state of calls. Instead, how about generating a new vector each team it is going to be sent out to the Plugins. It should be easy to filter on the State or whether it is conventional and Squelched.

taclane:
I know I've been really slow about getting all the changes merged back to the main repo, but we've pretty much found a way to handle this as a plugin-only solution.
- `calls_active()` can quickly iterate the call list and ignore conventional calls of zero length.
- The MQTT `client_id` string differentiates _distinct connections from a single trunk-recorder_ instance to the broker (if, for example someone was running both the mqtt_status_plugin *and* the mqtt_stats_plugin).
- The unused(?)`instanceId` can globally identify _which trunk-recorder_ is generating this information if multiple servers are aggregating to a single place.

PR #811, opened 2023-05-11
Strict CSV Parsing

robotastic:
What do people think about this? I found it lurking as a stale branch. Essentially it looks like it will exit if there are any malformed CSV lines. It is optional, turned with a config flag.


ZeroChaos-:
I am strongly for it but never got around to testing it, which is probably why it went stale.

Diego199069 (approved):

robotastic:
I made strict parsing the default with the updates to the TG / Channel file parsing

PR #816, opened 2023-05-21
update/edit README.md and docs/CONFIGURE.md, 3 minor fixes

rosecitytransit:
Includes adding a link to my decode rates logging plugin

3 minor fixes are:
-fix tbsk typo in p25_parser.cc
-remove duplicate include line in p25_frame_assembler_impl.cc
-remove duplicate add_file_log line in main.cc/have it capture log file config lines (fixes https://github.com/robotastic/trunk-recorder/pull/754)

Can someone test the plugin? I'm using code that I added directly to Trunk Recorder, and I'm wondering if the `override` keyword should be used. Also, is there a desire to be able to specify the name/location of the rates file?


lib/op25_repeater/lib/p25_frame_assembler_impl.cc
```diff
@@ -30,7 +30,6 @@
 #include <stdlib.h>
 #include <stdint.h>
 #include <string.h>
-#include <errno.h>
```
robotastic, line 33: Why are you removing this? OP25 has this included so it is easier to merge updates if I keep it in

trunk-recorder/main.cc
```diff
@@ -1756,15 +1758,6 @@ int main(int argc, char **argv) {
 
   start_plugins(sources, systems);
 
-  if (config.log_file) {
-    logging::add_file_log(
-        keywords::file_name = config.log_dir + "/%m-%d-%Y_%H%M_%2N.log",
-        keywords::format = "[%TimeStamp%] (%Severity%)   %Message%",
-        keywords::rotation_size = 100 * 1024 * 1024,
```
robotastic, line 1763: Do these lines need to be brought up? Looks like we are losing them

robotastic (commented):
Looks good! Just had 2 small questions

lib/op25_repeater/lib/p25_frame_assembler_impl.cc
rosecitytransit: Line 28 is identical to that one. Is there supposed to be two duplicate ones?

trunk-recorder/main.cc
rosecitytransit: these lines were duplicated at lines 161 and 162 in #754 and should have been deleted here

taclane:
Instead of calling out a specific plugin in .gitignore like `plugins/decode_rates`, can't we just leave it as `plugins/`so it could be used as staging area for other projects too?

rosecitytransit:
> Instead of calling out a specific plugin in .gitignore like `plugins/decode_rates`, can't we just leave it as `plugins/`so it could be used as a staging area for other projects too?

No, because the plugins folder is really for internal plugins that are included in the Trunk Recorder code, and should be managed by git. The official way to use an external plugin is to put the code somewhere else and run `sudo make install` so it's available as a system object. My way avoids needing to do that. But there could be an "eternal_plugins" folder that is ignored.

taclane:
Sorry, yeah.  You are correct.  All my plugins live in repos external to trunk-recorder and are manually installed.  

Maybe we can recommend `user_plugins\` directory as a staging area instead?  It's just going to be cumbersome if every potential plugin requires representation in .gitignore.

lib/op25_repeater/lib/p25_frame_assembler_impl.cc
robotastic: Ah! cool - it just didn't show up in the change review

trunk-recorder/main.cc
robotastic: Gotcha!

robotastic:
OK - I removed the plugin line from .gitignore. I would prefer not to have separate lines for each plugin there so I can minimize the number of PRs coming in. The best practice here would be to keep plugins outside of the Trunk Recorder repo. I should prob also clean out some of those weird directories from gitignore....

rosecitytransit:
> I would prefer not to have separate lines for each plugin there so I can minimize the number of PRs coming in.

Totally understood. Can you provide your thoughts/guidance on issue 482 (just see the last 2 comments)? Would like to work on that next.

PR #819, opened 2023-05-28
add option to disable output to console, docs/CONFIGURE.md formatting

rosecitytransit:
7c9766b creates a `consoleLog` config option to disable sending output to the console (e.g. for those who are using the `logFile` option), and moves the "Using Config file:" and "Version:" lines so they make it into the log file and not just the console

5b454f9 modifies the underlying formatting in docs/CONFIGURE.md (no actual text changes)


rosecitytransit:
@robotastic can you look at this and give your thoughts on issue 482 when you have a chance?

robotastic:
This looks good - Thanks!!!

I used a GUI tool to create the Table in the Config doc, so that probably screwed the original formatting up.

PR #820, opened 2023-05-29
P25 Trunking Message Improvements

robotastic:


rosecitytransit:
Anyway commit da9f918 can be reversed? The idea was to set repetitive system messages at `trace` level so that they don't clog up log files

PR #823, opened 2023-06-04
Feature/temp dir

robotastic:
Makes the directory used for Transmissions configurable.


rosecitytransit:
Not sure if you intended it, but just realized that in commit 1fdcafb (merged as 534a07d), the "Trunk Recorder" and "INSTANCE" log lines got moved up so that they output on the console even if `consoleLog` is `false`.

PR #824, opened 2023-06-04
correct callLog option in docs/CONFIGURE.md to true to match code

rosecitytransit:
I was going to change the default in the code to false, but then I realized that audioArchive defaults to true and am thinking callLog should match that. Otherwise, I should have included this in my other PR.

You want me to also add an entry for your tempDir option? I like your idea of having it in memory, it'll save hard drive wear, as well as address the issue of using a networked drive. The only downside being if trunk-recorder closes uncleanly, the current transmissions are lost.

But is it compatible with the transmissionArchive option for those who want to keep the files? How about making the option be "transmissionDir", with the net result of not needing to copy the finished file once complete?  I'm guessing the combined, uncompressed WAV file could be put in the transmissionDir since most people wouldn't want to keep that either.


robotastic:
ohhh.... good point, I think the current approach breaks the transmissionArchive option. I will go fix that. Of course I am not sure how many people use that feature. I will just move the Transmission Files to the captureDir if it is set. 

I am going to keep the Temp Dir for now, because a plugin may want to use it for other purposes too.

rosecitytransit:
> I will just move the Transmission Files to the captureDir if it is set.

My thought is if transmissionArchive or callLog is true, use the captureDir, but if not it's safe to use a tempDir. Or have the tempDir default be the captureDir, with a suggestion to change it to memory.

Also, if I wasn't clear, I meant just changing the name "tempDir" to "transmissionDir" since I think that's what it would be (though I guess it wouldn't have the /year/month/day structure, and if not, then I don't think it needs to be created each call: https://github.com/robotastic/trunk-recorder/blob/ee88a2d2baf6cf0c6cffe14e4d0fdeb37d48222d/trunk-recorder/gr_blocks/transmission_sink.cc#L83-L89)

rosecitytransit:
@robotastic since I see you created an entry for tempDir, you want to go ahead and merge this in? Also PR 482 should be ready to go.

robotastic:
Sorry - missed the comment in there. I think having a general TempDir could be useful in case a Plugin wants to use it for things beside transmissions.

PR #835, opened 2023-07-12
Include TDMA slot number in transmission_sink().

tadscottsmith:
Addresses an issue where P25P2 TDMA transmissions may not be correctly recorded if the same talk group and frequency are being utilized on different TDMA slots in short intervals.


robotastic:
It doesn't look like all the changes made it into the PR - I can take a look at this tonight and move the other parts in too. I might switch it around and just have Transmission sink pull the TDMA Slot from the Call that is passed.

tadscottsmith:
This was the the full change that a couple of people were using yesterday. Since you already had a transmission_sink::start_recording(Call *call, int slot) setup for DMR, it was just a matter of using the same function for the correct TDMA slot. It should also work to grab the slot within transmission sink and get rid of the overloaded star_recording() function.

robotastic:
Ahhh!!! Genius. That is what I get for trying to look at it before coffee.

PR #837, opened 2023-07-27
enable non free drivers for gr-osmosdr to get SDRplay for RSP1A support

robbiet480:
as requested by @gisforgirard on the Discord who also kindly donated hardware to me to validate this change


robbiet480:
Probably worth a docs update on the hardware page to point out Docker has native support for SDRPlay

robotastic:
I noted it on the README - but it doesn't look like the docs on Docker call out the SDRs that are supported. I should add that in...

PR #846, opened 2023-10-23
Capture P25 RFSS and Site ID information.

tadscottsmith:
Store P25 RFSS and Site ID in the system object and display them on initial decode. This should help with optimizing deduplication and other future enhancements.

### DIFF
--- a/trunk-recorder/main.cc
+++ b/trunk-recorder/main.cc
@@ -951,6 +951,14 @@ void current_system_status(TrunkMessage message, System *sys) {
   }
 }
 
+void current_system_sysid(TrunkMessage message, System *sys) {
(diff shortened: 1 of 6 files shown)


trunk-recorder/main.cc
```diff
@@ -951,6 +951,14 @@ void current_system_status(TrunkMessage message, System *sys) {
   }
 }
 
+void current_system_sysid(TrunkMessage message, System *sys) {
+  if ((sys->get_system_type() == "p25") || (sys->get_system_type() != "conventionalP25")) {
```
robotastic, line 955: Should this be != "conventionalP25"? Since this is an OR, it seems like it would be true for every system type, except for "conventionalP25

robotastic (commented):
Looks good - had a question on one of the checks

PR #852, opened 2023-11-09
Allow talkgroups to match a range of values

ArrEssJay:
This change allows talkgroups in `talkgroups.csv` to be defined as a range <min>:<max> as well as by a single value.

This is useful as a P25 phase 2 system that I receive has 'many' active talkgroups numbered 100-4000, all encrypted, for which I have no further details. Adding each one manually when it is discovered (phase 2 encrypted calls aren't flagged as such in trunking messages) is frustrating, and adding 4000 or so talkgroups, all with a name of 'unknown' and a priority of -1 seems rather inelegant. 

This change seeks to remedy this. I had considered adding wildcard/regex matching, but this seems like the simplest solution.

### DIFF
--- a/docs/CONFIGURE.md
+++ b/docs/CONFIGURE.md
@@ -411,22 +411,28 @@ The matching simplestream config to send audio from talkgroup 58918 to TCP port 
 
 This file provides info on the different talkgroups in a trunking system. A lot of this info can be found on the [Radio Reference](http://www.radioreference.com/) website. You need to be a Radio Reference member to download the table for your system preformatted as a CSV file. You can also try clicking on the "List All in one table" link, selecting everything in the table and copying it into a spreadsheet program, and then exporting or saving as a CSV file.
 
-**Note:** You can use the direct CSV from Radio Reference for talk groups, but setting priority is not supported with this file format.  If you need to use the Priority field, you'll need to reorder the CSV to match the format described below.
-
 You may add an additional column that adds a priority for each talkgroup. The priority field specifies the number of recorders the system must have available to record a new call for the talkgroup. For example, a priority of 1, the highest means as long as at least a single recorder is available, the system will record the new call. If the priority is 2, the system would at least 2 free recorders to record the new call, and so on. If there is no priority set for a talkgroup entry, a prioity of 1 is assumed.
+
+Talkgroups matching a range of decimal numbers may also be defined. In this case the channel numbers should take the form <min>:<max>. The hex
+value is not used. Note that this formatting will not be detected if the radio reference style. This is useful when receiving a system with a range of talkgroups in use by a agency using encryption, for which no further details are available. This is doubly problematic for P25 phase 2 talkgroups, where the trunking system does not flag calls as encrypted before Trunk Recorder attempts to record them. Rather than enter every unknown talkgroup as it is discovered, the entire range can be set to a priority of -1 and ignored.
+
+When Trunk Recorder looks up a talkgroup, it does so by its decimal number. The first talkgroup in the list with a single matching number will be used. If there is no match, then the first talkgroup specified by a range that includes the supplied talkgroup number will be returned, if one exists.
 
 Talkgroups assigned a priority of -1 will never be recorded, regardless of the number of available recorders.
 
 Trunk Recorder really only uses the priority information and the decimal talkgroup ID. The Website uses the same file though to help display information about each talkgroup.
 
+**Note:** You can use the direct CSV from Radio Reference for talk groups, but talkgroup priority and talkgroup ranges are not supported with this file format. If you need to use either or these features, you'll need to reorder the CSV to match the format described below.
+
 Here are the column headers and some sample data: NOTE: If you are adding the Priority to a RR csv, as well as changing order you must use a heading for the first column other than "Decimal" eg DEC for TR to detect you are supplying this layout.
 
-| DEC | HEX | Mode | Alpha Tag    | Description    | Tag            | Group    | Priority | PreferredNAC (optional) |
-|-----|-----|------|--------------|----------------|----------------|----------|----------|-------------------------|
-|101  | 065 | D    | DCFD 01 Disp | 01 Dispatch    | Fire Dispatch  | Fire     | 1        | 1000                    |
-|2227 | 8b3 | D    | DC StcarYard | Streetcar Yard | Transportation | Services | 3        | 1001                    |
+| DEC  | HEX | Mode | Alpha Tag    | Description    | Tag            | Group    | Priority | PreferredNAC (optional) |
+|------|-----|------|--------------|----------------|----------------|----------|----------|-------------------------|
+|101   | 065 | D    | DCFD 01 Disp | 01 Dispatch    | Fire Dispatch  | Fire     | 1        | 1000                    |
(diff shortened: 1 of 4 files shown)


robotastic:
Let me go dig into this more... it seems like Phase 2 should be flagging encrypted message in the control stream. Maybe I am not processing something correctly. It seems like a huge pain to have to flag all the encrypted talkgroups manually.

ArrEssJay:
My reading of the P25 specs is that trunking messages do not flag encryption state for phase 2. I can dig out the reference if that helps. It's certainly what I've observed. I did have a look at getting the flag out of the recorder, but this currently means setting up a recorder just to find out if its something we want to record or not.

robotastic:
Could you try setting the `newCallFromUpdate` setting to `false` in config.json?  What is prob happening is that those Calls are being started off UPDATE messages, which do not have the encryption flag set in them.

<!DOCTYPE html>

newCallFromUpdate |   | true | true / false | Allow for UPDATE trunking messages to start a new Call, in addition to GRANT messages. This may result in more Calls with no transmisions, and use more Recorders. The flipside is that it may catch parts of a Call that would have otherwise been missed. Turn this off if you are running out of Recorders.
-- | -- | -- | -- | --

ArrEssJay:
I'll check that. Phase 2 traffic is not being flagged as encrypted in trunking messages. I remember reading somewhere that this was normal, but I've not been able to track down the source, nor can I find that section of the spec openly available on the web.

Relevant log messages:

```
023-11-13 11:33:51.833894] (debug)   [ 0]tsbk02	MOTOROLA_OSP_PATCH_GROUP_CHANNEL_GRANT	Channel ID: 53292	Freq: 165.462500 MHz	sg    3423	TDMA 0	sa 320717
[2023-11-13 11:33:51.834031] (info)   [allambee-353]	1C	TG:       3423 (                      -)	Freq: 165.462500 MHz	TG not in Talkgroup File
[2023-11-13 11:33:51.834091] (info)   [allambee-353]	1C	TG:       3423 (                      -)	Freq: 165.462500 MHz	Starting P25 Recorder Num [0]	TDMA: true	Slot: 0	QPSK: true
[2023-11-13 11:33:52.111779] (debug)   [ 0]tsbk30 TDMA SYNCHRONIZATION BROADCAST
[2023-11-13 11:33:52.111942] (debug)   [ 0]tsbk20	Acknowledge Response	ga 65533	sa 3115382	Reserved: 255
[2023-11-13 11:33:52.112048] (debug)   [ 0]tsbk20	Acknowledge Response	ga 65533	sa 3112801	Reserved: 255
11/13/23 11:33:52.112439 MAC_PTT: srcaddr=320717, grpaddr=3423, algid=84, keyid=5d, mi=fd 87 5c 34 1a 16 51 21 00, rs_errs=0
11/13/23 11:33:52.112754 MAC_PTT: srcaddr=320717, grpaddr=3423, algid=84, keyid=5d, mi=fd 87 5c 34 1a 16 51 21 00, rs_errs=0
11/13/23 11:33:52.287761 [0] 4V_BURST(0) ESS: algid=80, keyid=0, mi=00 00 00 00 00 00 00 00 00, rs_errs=0
```

ArrEssJay:
I tried adding parsing of voice service options for `MOTOROLA_OSP_PATCH_GROUP_CHANNEL_GRANT`:

```
     unsigned long opts  = bitset_shift_mask(tsbk, 72, 0xff);
        bool emergency = (bool)bitset_shift_mask(tsbk, 72, 0x80);
        bool encrypted = (bool)bitset_shift_mask(tsbk, 72, 0x40);
        bool duplex = (bool)bitset_shift_mask(tsbk, 72, 0x20);
        bool mode = (bool)bitset_shift_mask(tsbk, 72, 0x10);
        int priority = bitset_shift_mask(tsbk, 72, 0x07);
        
        unsigned long ch = bitset_shift_mask(tsbk, 56, 0xffff);
        unsigned long sg = bitset_shift_mask(tsbk, 40, 0xffff);
        unsigned long sa = bitset_shift_mask(tsbk, 16, 0xffffff);
        unsigned long f = channel_id_to_frequency(ch, sys_num);

        message.message_type = GRANT;
        message.freq = f;
        message.talkgroup = sg;
        message.source = sa;
        
        message.encrypted = encrypted;
        message.emergency = emergency;
        message.duplex = duplex;
        message.mode = mode;
        message.priority = priority;

      if (get_tdma_slot(ch, sys_num) >= 0) {
        message.phase2_tdma = true;
        message.tdma_slot = get_tdma_slot(ch, sys_num);
      } else {
        message.phase2_tdma = false;
        message.tdma_slot = 0;
      }

      os << "tsbk02\tMOTOROLA_OSP_PATCH_GROUP_CHANNEL_GRANT\tChannel ID: " << std::setw(5) << ch << "\tFreq: " << format_freq(f) << "\tsg " << std::setw(7) << sg << "\tTDMA " << get_tdma_slot(ch, sys_num) << "\tsa " << sa;
      message.meta = os.str();
      BOOST_LOG_TRIVIAL(debug) << os.str();
```

These calls are now correctly flagged as encrypted.

```
[2023-11-13 12:00:42.266377] (debug)   [ 0]tsbk02	MOTOROLA_OSP_PATCH_GROUP_CHANNEL_GRANT	Channel ID: 53316	Freq: 165.612500 MHz	sg    3423	TDMA 0	sa 3112062
[2023-11-13 12:00:42.273226] (info)   [allambee-353]	1C	TG:       3423 (                      -)	Freq: 165.612500 MHz	Not Recording: ENCRYPTED - src: 3112062
```

Somewhat confusing, I appreciate, given its a manufacturer specific variant of an `update` message.

ArrEssJay:
And as expected `MOTOROLA_OSP_PATCH_GROUP_CHANNEL_GRANT_UPDATE` messages do not contain service options, in which case yes, calls being started from updates would be a problem. Unfortunately I have issues with co-channel interference and do drop trunking messages from some transmission sites, so this isn't ideal.

robotastic:
I like that idea of just having a recorder Stop when it receives encrypted - the call would have to stick around and change from being in Record mode to Monitor mode, otherwise it would just start back up again when the next UPDATE message comes through.

One other thought - how about setting the **recordUknown** key to *false* in the system section of the config file? That way only the defined talkgroups will be recorded. The only downside is that you would miss any new talkgroups that pop up. 

<!DOCTYPE html>

recordUnknown |   | true | true / false | Record talkgroups if they are not listed in the Talkgroups File.
-- | -- | -- | -- | --

Ultimately adding support for ranges might be the best option... I am just trying to keep the code base simple as possible and seeing if it is possible to address the root problem.

robotastic:
Could you add a PR for handling MOTOROLA_OSP_PATCH_GROUP_CHANNEL_GRANT? Nice detective work!

ArrEssJay:
https://github.com/robotastic/trunk-recorder/pull/869 for `MOTOROLA_OSP_PATCH_GROUP_CHANNEL_GRANT`

It's straightforward to check whether encryption can/does succeed or not by checking the return state of `crypt_algs.prepare()`. I'm just taking a look to see if there's a straightforward/reliable way to pass a message back up the stack to the recorder to abort the call if we know it's going to fail.

robotastic:
Sorry for the delay on this - this seems like a good feature to no matter what TR should be doing with encryption. Since this PR came in, the handling for CSV files changed around. I will try to rebase it with the new approach

PR #854, opened 2023-11-11
README: add Prometheus plugin

USA-RedDragon:
I created a Prometheus metrics plugin at https://github.com/USA-RedDragon/trunk-recorder-prometheus, so I wanted to link it here. If this isn't something you want to link, feel free to close the PR, no hard feelings :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,6 +74,7 @@ By default, Trunk Recorder just dumps a lot of recorded files into a directory. 
 * [MQTT Statistics](https://github.com/robotastic/trunk-recorder-mqtt-statistics): Publishes statistics about a Trunk Recorder instance over MQTT
 * [Decode rates logger](https://github.com/rosecitytransit/trunk-recorder-decode-rate): Logs trunking control channel decode rates to a CSV file, and includes a PHP file that outputs an SVG graph
 * [Daily call log and live Web page](https://github.com/rosecitytransit/trunk-recorder-daily-log): Creates a daily log of calls (instead of just individual JSON files) and includes an updating PHP Web page w/audio player
+* [Prometheus exporter](https://github.com/USA-RedDragon/trunk-recorder-prometheus): Publishes statistics to a metrics endpoint via HTTP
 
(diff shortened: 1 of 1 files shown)


robotastic:
This is great!! I am going to go give this a try - it seems like a super easy way to track stats.

frameshift18:
Trying to get this to work and ran into a couple issues.

1) In trunk-recorder/plugin_manager/plugin_api.h I had to change #include <json.hpp> to #include "../json.hpp" or else it couldn't find it. I know you just changed that so probably a typo?

2) When trying to build the plugin I get the following error and I don't know enough about C to fix it, lol :D. I did remove 'override' and it built but it doesn't seem to work then. At least nothing loads at the :9842 endpoint and Prometheus doesn't find any metrics.

`prometheus_plugin.cc:60:7: error: ‘int Prometheus::parse_config(boost::property_tree::ptree&)’ marked ‘override’, but does not override
   60 |   int parse_config(boost::property_tree::ptree &cfg) override`

USA-RedDragon:
Yeah the plugin was built against 4.2.0, so the new `json` stuff being passed around in the latest changes on the `master` branch aren't used yet since the API changed a bit. Getting my plugin's main based on the `master` is on my todo tonight. There now exists a 4.2.0 tag on the GHCR image which should run against Trunk Recorder 4.2.0.

taclane:
> Trying to get this to work and ran into a couple issues.
> 

A very recent change to trunk recorder (just today) changed how plugins access information from the config.json. 

This plugin should work fine if you pick a slightly earlier commit or use one of the release tarballs, but authors will need a moment to catch up to some breaking changes if their plugins are going to built against a git pull. 

Both 1 and 2 should be explained by commits that just got pushed out.

frameshift18:
Kind of figured. Course I had to update my running TR instance to master and then try this. All good and appreciate the work :)

robotastic:
Sorry about that - I was going to put in a PR to fix things.

PR #857, opened 2023-11-12
Fix channel CSV parsing, allow for freqs in MHz, other config parsing fixes

EricTendian:
It seems like there was a copy-paste error in the new talkgroup/channel CSV parsing, so the channel CSV was incorrectly expecting a Decimal column header. This fixes it to the right column header.

Additionally, the new logic expects a float for frequency, which is backwards-incompatible with the frequencies in existing CSVs. To work around this, while also supporting the RadioReference format of MHz freqs, it will instead check if it's either an int or a float, and if it's a float (and under 1000), will convert the MHz frequency to Hz.

Third, the config parser expected the system IDs for Boadcastify and Rdio Uploader to be strings when they really should be ints (and were previously). This updates the actual data types to be int, so there are no type issues.


robotastic:
Rolled into #862

Issue #858, opened 2023-11-12
Broadcastify System ID string/number config.json issue


james-coder:
I've recently compiled trunk-recorder on a WSL2 Ubuntu 20.04 system. (Using usbipd.)

I used a previously working config.json from another computer where `"broadcastifySystemId": "86753"` (Notice the quotes around the number.) This was working fine for me with no errors.

When I compiled it for the WSL2 Ubuntu 20.04 system, I get the error:
```
...
-------------------------------------
SYSTEMS
-------------------------------------
...
[2023-11-11 21:12:58.423169] (info)   Broadcastify API Key: BROADCASTIFYAPIKEY
[2023-11-11 21:12:58.423216] (error)   Failed parsing Config: [json.exception.type_error.302] type must be number, but is string
```
Which leads me to think it wants a int there instead instead of a string. I change it to be `"broadcastifySystemId": 86753` , but then I get this error:
```
...
-------------------------------------
PLUGINS
-------------------------------------
...
[2023-11-11 21:18:04.382722] (info)   Setting up plugin -  Name: broadcastify_uploader	 Library file: libbroadcastify_uploader.so
[2023-11-11 21:18:04.383107] (info)   Broadcastify Server: https://api.broadcastify.com/call-upload
[2023-11-11 21:18:04.383222] (error)   Failed parsing Config: [json.exception.type_error.302] type must be string, but is number
```

So basically, I getting one error message that says it should be an integer (in SYSTEMS), I change it to an integer, then I get another error message saying it should be a string (in PLUGINS).

The documentation at https://github.com/robotastic/trunk-recorder/blob/master/docs/CONFIGURE.md#system-object says:

**broadcastifySystemId**: if broadcastifyCallsServer is set System ID for Broadcastify Calls (**this is an integer**, and different from the RadioReference system ID)

The broadcastify plugins code is in https://github.com/robotastic/trunk-recorder/blob/47460b1582ec5187ed99eac0df93ae41ab93017f/plugins/broadcastify_uploader/broadcastify_uploader.cc

james-coder:
Here is my config.json:
```
{
    "ver": 2,
    "defaultMode": "analog",
    "sources": [
        {
            "center": 123456789,
            "rate": 20000000,
            "error": 0,
            "gain": 0,
            "ifGain": 24,
            "bbGain": 28,
            "digitalRecorders": 3,
            "analogRecorders": 8,
            "debugRecorders": 0,
            "driver": "osmosdr",
[...]

PR #859, opened 2023-11-12
Broadcastify Plugin: System ID should be an int

james-coder:
See https://github.com/robotastic/trunk-recorder/issues/858

I've tested this resolves the issue pointed out in the above issue.


robotastic:
Good catch - thanks!

PR #860, opened 2023-11-12
rdio-scanner: Fix systemID to match the datatype used in rdio-scanner

USA-RedDragon:
rdio-scanner uses a Go uint (uint32_t is the C equivalent) to store the system ID: https://github.com/chuot/rdio-scanner/blob/c9c4ea4530f906dc2f1a9d67706ca987e5f82131/server/system.go#L29

This also prevents https://github.com/robotastic/trunk-recorder/commit/47460b1582ec5187ed99eac0df93ae41ab93017f from making this doc wrong, as the current state of the code requires `systemID` to be a string: https://github.com/robotastic/trunk-recorder/blob/713049ad540423392db5ed8b7fba3507bb0c0fea/docs/Plugins.md?plain=1#L50
https://github.com/robotastic/trunk-recorder/blob/713049ad540423392db5ed8b7fba3507bb0c0fea/docs/Plugins.md?plain=1#L66

### DIFF
--- a/plugins/rdioscanner_uploader/rdioscanner_uploader.cc
+++ b/plugins/rdioscanner_uploader/rdioscanner_uploader.cc
@@ -14,7 +14,7 @@
 struct Rdio_Scanner_System {
   std::string api_key;
   std::string short_name;
-  std::string system_id;
+  uint32_t    system_id;
   std::string talkgroupsFile;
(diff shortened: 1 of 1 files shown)


robotastic:
Good catch!

PR #861, opened 2023-11-12
ci: push to both GHCR and Docker Hub

USA-RedDragon:
GHCR is free and it give you a redundant source for OCI images in the event that Docker Hub has an outage. No work required on your part, due to the `jobs.buildx.permissions.packages` write permission. `contents: read` is the default, but would be overridden if not present. Merging this would create `ghcr.io/robotastic/trunk-recorder`

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -12,6 +12,9 @@ on:
 jobs:
   buildx:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


robotastic:
Great idea!! I think some of my builds were failing yesterday because there were some sort of Docker Hub issues.

PR #863, opened 2023-11-13
Multisite updates and improvements

taclane:
Building on the improvements in #846, trunk-recorder should use RFSS and Site ID as the default discriminators when multisite is enabled for P25 systems.  This overcomes an issue where multiple P25 sites broadcasting the same NAC can circumvent de-duplication logic.

Prior to this, some P25 systems required manual setup using a `multiSiteSystemName`, and it may not have been apparent why it did not appear to be working using the default settings.

The existing "Preferred NAC" or `multiSiteSystemNumber` functionality has not changed, but users may additionally use a decimal RFSS/SiteID (in the form `RRRRssss`, e.g. `00030013`) to match talkgroups to preferred sites.

The console message has been updated to better show the "WACN/RFSS-Site" match when it occurs:
```
Duplicate Grant - 20C: BEE00/3-13 23C: BEE00/3-14  Source: 1819332 Call: 20C State: recording
```
Manually defining a multisite system with `multiSiteSystemName`or `multiSiteSystemNumber` remains unchanged, but the log message would look as below:
```
Duplicate Grant - 66C: CLMRN/0 67C: CLMRN/1  Source: 1849171 Call: 66C State: recording
```

### DIFF
--- a/docs/CONFIGURE.md
+++ b/docs/CONFIGURE.md
@@ -220,9 +220,9 @@ There is a list of available Plugins [here](./Plugins.md).
 
 When enabled, Multi-Site mode attempts to avoid recording duplicate calls by detecting simulcasted transmissions for the same talkgroup across multiple sites at the same time.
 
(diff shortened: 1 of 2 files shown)


robotastic:
This all sort of makes sense to me... but the code looks good, let's ship it!

PR #865, opened 2023-11-14
load channelFile items into system channels vector

taclane:
When a `channelFile` is used for conventional systems, these channels are successfully created and monitored, but are not added to the system `channels` vector.  Downstream plugins or other API calls may want to access a list of conventional channels assigned to a system with `sys->get_channels()` whether or not a channelFile is in use.

This change does not affect the methods trunk-recorder uses to set up conventional systems with a channelFile, it only loads those channels into the internal vector after the .csv file has been read.


robotastic:
Great addition... I feel like we could also probably simplify this section of code now, since conventional talkgroups will now be loaded into the channels vector. I will go check it out tonight:
https://github.com/robotastic/trunk-recorder/blob/34484fe28a11c481d72c905636beded444b37c0c/trunk-recorder/main.cc#L1023-L1037

PR #866, opened 2023-11-14
improve duplicate detection for p25p2 TDMA

taclane:
Multi-site duplicate detection currently requires _both_ calls to have the same TDMA slot number, even across different systems or frequencies.  Depending on the slot assignments from the system controllers, there is only a 50% chance to correctly identify duplicate calls across TDMA networks.

Removing the line below requiring the same TDMA status and slot essentially doubles the reliability of dupe detection on these systems.

```cpp
if ((call->get_phase2_tdma() == message.phase2_tdma) && (call->get_tdma_slot() == message.tdma_slot)) {
```

### DIFF
--- a/trunk-recorder/main.cc
+++ b/trunk-recorder/main.cc
@@ -486,55 +486,54 @@ void handle_call_grant(TrunkMessage message, System *sys, bool grant_message) {
     Call *call = *it;
 
     /* This is for Multi-Site support */
+    // Find candidate dublicate calls with the same talkgroup and different multisite-enabled systems
     if (call->get_talkgroup() == message.talkgroup) {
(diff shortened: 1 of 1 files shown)


trunk-recorder/main.cc
```diff
@@ -486,55 +486,54 @@ void handle_call_grant(TrunkMessage message, System *sys, bool grant_message) {
     Call *call = *it;
 
     /* This is for Multi-Site support */
+    // Find candidate dublicate calls with the same talkgroup and different multisite-enabled systems
```
EricTendian, line 489:
```suggestion
    // Find candidate duplicate calls with the same talkgroup and different multisite-enabled systems
```
Just fixing a typo.

EricTendian (changes requested):

EricTendian (approved):

taclane:
Streamlined the console log entry per Discord feedback.

PR #867, opened 2023-11-14
call_info update

taclane:
The `Call_Data_t` struct currently does not contain the internal system number on which a call occurred.  Should plugins wish to use API functions to look up additional information related to a completed call, they cannot do so without some form of "reverse lookup" by `shortName` which is not guaranteed to be a unique identifier.

Including the internal system number in completed call data would allow direct and unique access to related system information through a simple numerical comparison.


robotastic:
That makes sense - good add

PR #869, opened 2023-11-14
Parse Motorola Patch Group Grant service options

ArrEssJay:
Service options (encryption etc.) for Motorola Patch Group Grants were not being parsed. Added this. They are not present in the update version. The patch service options already commented out relate to the 'moto patch group add' message so I removed these lines completely.

Just taking another look at this code, we parse the Patch Group as being equivalent to a Talkgroup, when it's specifying that it should also be patched to the target talkgroups listed for that patch group. We could also get a call to a talkgroup with the same number that should not be patched, unlikely as this seems.

I do sometimes see some inconsistent behaviour with regards to patched call handling and the talkgroup that trunk recorder lists as 'the' talkgroup that it was on, and talkgroup mismatch errors between the trunking/recorder.

Hence it may be worth being more diligent in the way these are tracked.

### DIFF
--- a/trunk-recorder/systems/p25_parser.cc
+++ b/trunk-recorder/systems/p25_parser.cc
@@ -343,15 +343,28 @@ std::vector<TrunkMessage> P25Parser::decode_tsbk(boost::dynamic_bitset<> &tsbk, 
     // Options are not present in an UPDATE
 
     if (mfrid == 0x90) {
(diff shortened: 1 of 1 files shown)


robotastic:
Thanks - that looks good, merging it in

I am confused though - this same code gets called for GRANTS and PATCHES?

PR #870, opened 2023-11-15
better uint8_t fix

taclane:
Adjusting the type causes more problems than adjusting the error message.  Downstream effects can be bad NACs and XOR masks.


robotastic:
It was worth a shot!

Issue #875, opened 2023-11-26
Blank recordings and duplicates uploaded to Broadcastify. SOX warning.


gonzoblue:
During the normal running I will see something like this go by:

[2023-11-22 15:30:54.005379] (error)   [SD700]  5C      TG:        101 (             WES DISP 1)         Freq: 769.756250 MHz   No Transmissions were recorded!
[2023-11-22 15:30:54.005437] (error)   [SD700]  6C      TG:       8192 (        SD Sanitation 2)         Freq: 769.318750 MHz   No Transmissions were recorded!
[2023-11-22 15:30:56.018835] (error)   [SD700]  14C     TG:        323 (               ADMIN 7A)        Freq: 771.068750 MHz    Dropping 1440 samples - current_call is null     Rec State: available   Since close: 1
[2023-11-22 15:30:57.010825] (error)   [SD700]  17C     TG:       8192 (        SD Sanitation 2)         Freq: 769.318750 MHz   No Transmissions were recorded!
[2023-11-22 15:31:05.009238] (error)   [SD700]  59C     TG:       1621 (                 6B SLJ)         Freq: 769.818750 MHz   No Transmissions were recorded!
[2023-11-22 15:31:05.010745] (error)   [SD700]  60C     TG:       8192 (        SD Sanitation 2)         Freq: 769.318750 MHz   No Transmissions were recorded!
[2023-11-22 15:31:08.005543] (error)   [SD700]  68C     TG:        407 (                 CMD 7H)         Freq: 771.268750 MHz   No Transmissions were recorded!
[2023-11-22 15:31:08.006323] (error)   Somehow, /dev/shm/SD700/101-1700695867_770256250.wav doesn't exist, not attempting to provide it to sox
sox:      SoX v14.4.2
sox FAIL sox: Not enough input filenames specified

And then I get a lot of these messages, and when checking my recordings on Broadcastify they are all 0 seconds in length and basically contain nothing:

Broadcastify Upload Skipped: SKIPPED---ALREADY-RECEIVED-THIS-CALL

I have also seen a message about sox and fdkaac are required, but both are installed.
This happens as of the latest git pull on 11/22/23. I have been running and older version without issue for a month or two.
For now I have checked out the version from 40 days ago (I picked a slightly random number of days when I might have last updated). It is back to working just fine.

gonzoblue:
[...]

PR #876, opened 2023-11-28
Adjust integer config values to double

taclane:
The value for `squelch_db` was previously read from the config file as a double, and is later used as a double during recorder setup.  Updating the default value would restore this behavior, and avoid truncating downward to an integer.

The value for `callTimeout` has previously been read from the config file as an integer, but the json library is now truncating the value instead of rounding.  Reports such as #875, and similar discussions on the discord, seem to indicate that users have been setting decimal values for this under one second, and the fact that it is now truncating down to 0 is causing multiple unexpected issues.

Much like other call-shaping parameters (`minDuration`, `maxDuration`, and `minTransmissionDuration`), `callTimeout` should also allow decimal values.  Adjusting the type in the config struct would correct this.


robotastic:
Agreed - the difference between rounding and truncating has seemed to cause some real issues. This is a good approach to address that.

PR #877, opened 2023-11-28
Demote "Decoding System Site" error message

taclane:
This log message shown on startup is not an error, and should be set to the info level.


robotastic:
Agreed - this mostly happens because it simply hasn't been received yet.

PR #878, opened 2023-11-28
Proposed Feature: NO_COLOR console or logfile output

taclane:
Trunk Recorder extensively uses ANSI color codes in logging statements to make the console output easier to read, but users should be able to disable them if desired.  

Additionally, this behavior can create issues when combined with the `logFile` option, as the raw escape sequences may make the output difficult to parse or ingest (example below).
```
[2023-11-28 13:47:07.413195] (info)   [clmrn]   ^[[0;34m10C^[[0m        TG:      23507 (^[[35m            Stamford FD^[[0m)     Freq: 769.531250 MHz    ^[[33mConcluding Recorded Call^[[0m - Last Update: 0s   Recorder last write:0.158642    Call Elapsed: 2
[2023-11-28 13:47:07.414909] (info)   [clmrn]   ^[[0;34m10C^[[0m        TG:      23507 (^[[35m            Stamford FD^[[0m)     Freq: 769.531250 MHz    ^[[33mCall was UPDATE not GRANT^[[0m
[2023-11-28 13:47:07.416419] (info)   [clmrn]   ^[[0;34m10C^[[0m        TG:      23507 (^[[35m            Stamford FD^[[0m)     Freq: 769.531250 MHz    ^[[33mStopping P25 Recorder Num [0]^[[0m        TDMA: false     Slot: 0 Hz Error: 96
```
Proposed is a `logColor` setting to selectively disable color output to logfiles and/or the console. 

- By default, color codes will remain enabled for console output, but now get removed from any configured `logFile`.  
- This default will be overridden by the presence of a [`NO_COLOR` environment variable](https://no-color.org).  If the environment variable is set (and not NULL), it will disable color output to both the console and logfiles without needing to alter the config.json.
- If `logColor` is used in the config.json, that value will override the presence of the `NO_COLOR` env var.  Per no-color.org recommendations, this is the desired behavior.
   > **How should configuration files and command-line arguments be processed in the presence of $NO_COLOR?**
   >     User-level configuration files and per-instance command-line arguments should override $NO_COLOR.

Attached below is an example of the default setting, showing the simultaneous output of the console and logfiles.  The color codes have been automatically removed from the logfiles without the use of additional tools or commands.
logs


robotastic:
RegEx is dark magic and this is an amazing addition... thank you!

Issue #843, opened 2023-08-29
Fatal crash: boost "Invalid cross-device link"


PhilRW:
* `Linux <hostname-redacted> 5.10.0-25-amd64 #1 SMP Debian 5.10.191-1 (2023-08-16) x86_64 GNU/Linux`
* `Docker version 24.0.5, build ced0996`
* Using `edge` docker image from docker hub
* /data is mounted volume

 Program crashes with the following:
 ```
 boost::filesystem::copy_file: Invalid cross-device link: "/dev/shm/codtrs5/9580-1693338509_852987500.wav", "/data/codtrs5/2023/8/29/9580-1693338509_852987500.wav"
0x7f443a0325d9: (gr::tagged_stream_block::check_topology(int, int)+0x2e49)
0x7f4439c2f24c: (std::rethrow_exception(std::__exception_ptr::exception_ptr)+0x7c)
0x7f4439c2f2b7: (std::terminate()+0x17)
0x7f4439c2f23e: (std::rethrow_exception(std::__exception_ptr::exception_ptr)+0x6e)
0x5595a221e941: (Call_Concluder::manage_call_data_workers()+0xeb1)
0x5595a2140604: (monitor_messages()+0x394)
0x5595a2134210: (main+0x740)
0x7f443987bd90: (__libc_init_first+0x90)
0x7f443987be40: (__libc_start_main+0x80)
0x5595a2137ab5: (_start+0x25)
```

PhilRW:
Problem seems to be mitigated by setting `transmissionArchive` to `false`.

taclane:
If you still want to keep transmission archives, the other option is to set `tempDir` to the same directory (or at least drive) as `captureDir` in the config file.  Keeping both of those on the same device should avoid the issue, but you'll miss any benefit of recording all the individual transmissions to a tempfs instead of storage media.

There are a handful of boost library/kernel combos that can cause this, but it's ultimately related to a kernel issue that existed between linux 5.3 and 5.18.  Boost created a workaround at some point, and it was fixed in the 6.x kernel, but some distros like debian 11 might still run into the "cross-device link" error.

sally-yachts:
Just for posterity's sake I'd like to confirm taclane's findings. My main recorder ran the TR official docker image on a Debian 11 box with a backported 6.x kernel and still ran into this error. It was configured to archive transmissions and tempDir wasn't set - I configured it to use a directory on the same volume as the existing audio storage and I can now run newer code without problem.

[...]

PR #886, opened 2023-12-05
Fs copy fix for "Invalid cross-device link"

taclane:
Sometime between linux 5.3 and 5.18, a regression prevented files from copying across devices using `copy_file_range()`.  Boost created a workaround for 1.78, and it was fixed in the 6.x kernel, but depending on what packages are offered in the distro, this problem may still remain.

Replicated with the following system:
```
libboost-all-dev/mantic,now 1.74.0.3ubuntu7 amd64 [installed]
libboost-filesystem-dev/mantic,now 1.74.0.3ubuntu7 amd64 [installed,automatic]

Linux system2 6.5.0-13-generic #13-Ubuntu SMP PREEMPT_DYNAMIC Fri Nov  3 12:16:05 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux
```
A `boost::filesystem::copy_file: Invalid cross-device link:` will occur when `transmissionArchive` is true, and trunk-recorder attempts to copy the tx wavs from a ram-based `tempDir` to conventional storage at `captureDir` at the conclusion of a call.  Previous mitigation techniques were either to disable transmissionArchive, or move the tempDir to the same device as captureDir.

This should help address issues like #843 by allowing the users with affected kernels to manually install newer version of boost libs, or allow users with 6.x kernels and affected boost libs to use std::filesystem instead.

### DIFF
--- a/trunk-recorder/call_concluder/call_concluder.cc
+++ b/trunk-recorder/call_concluder/call_concluder.cc
@@ -143,13 +143,15 @@ void remove_call_files(Call_Data_t call_info) {
       // if the files are being archived, move them to the capture directory
       for (std::vector<Transmission>::iterator it = call_info.transmission_list.begin(); it != call_info.transmission_list.end(); ++it) {
         Transmission t = *it;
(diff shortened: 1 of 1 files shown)


robotastic:
You are a saint for hunting down the root cause of this bug and putting together a patch!

PR #887, opened 2023-12-07
make it work with boost 1.78+

taclane:
Compile fix for boost 1.78+.


robotastic:
Thanks again for patching this up!

PR #888, opened 2023-12-21
Add TSBK Stream Option

jelimoore:
This PR adds the ability to send TSBKs to an external UDP host for further analysis. Simply add the following lines to your config in the main section:

```"tsbkStreamServer": "1.2.3.4",
"tsbkStreamPort": 5678```

And you will get all incoming TSBKs (single-block and multi-block) sent over the UDP socket to the host and port of your choice.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -233,6 +233,7 @@ list(APPEND trunk_recorder_sources
   trunk-recorder/formatter.cc
   trunk-recorder/source.cc
   trunk-recorder/call_conventional.cc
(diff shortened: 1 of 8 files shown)


robotastic:
Sorry for the delay on this - The plugin system has the ability to export trunking message via the `plugman_trunk_message` I would prefer to keep it destination agnostic like this. A plugin could use the information to either send it to a UDP port or write to a DB or something else fun.

If the plugin should be getting more raw info or different details, happy to investigate that further.

PR #891, opened 2023-12-30
simplestream updates to send variable-length JSON metadata along with streaming audio packets

aaknitt:
Added 'sendJSON' configuration to simplestream plugin.

When set to true, JSON metadata that contains talkgroup, freq, src, patched_talkgroups, and audio_sample_rate fields will be prepended to each audio packet.  Length of the JSON metadata will be prepended (4 bytes, little endian) to the metadata.  

sendJSON will override the older sendTGID parameter (deprecated).

example_audio_player.py updated to support receiving & acting on JSON metadata.


robotastic:
Looks good!

PR #896, opened 2024-01-06
trace log some repetitive P25 messages

rosecitytransit:
looks like I actually set everything to debug in e4f843f; shouldn't have

also debug log only new patches


rosecitytransit:
if we're not doing it, may also want to add current patch listing to the print status

robotastic:
I like it - let me see how to add Patches to status message

rosecitytransit:
Oh, and if you do that, remove it from the Decoding Rate messages and maybe have the status messages print more frequently, say every 60 or 120 seconds

PR #905, opened 2024-01-15
Fix minor typos in debug logging for opcode 0x28

anthonykava:
Fix minor typos in debug logging for opcode 0x28


robotastic:
Looks good - thanks for the fix!

PR #907, opened 2024-01-21
OpenMHz upload improvements

taclane:
The OpenMHz plugin currently only uploads to the same system as a trunk-recorder `shortName`.   

This modification to the plugin will allow trunk-recorder systems to upload to an arbitrary OpenMHz endpoint that you have an API key for, even when it does not match the configured `shortName`. 

Primarily, this better enables multiple systems to contribute to the same OpenMHz feed.  This is especially useful for multi-site systems, where the only way for these setups to work with OpenMHz was to give them all the same `shortName` in trunk-recorder.  While it is an effective workaround, it can be undesirable from the standpoint of logging and diagnostics for multiple radio systems to report with the same name.

Additionally, the plugin startup messages are improved to more clearly reflect which systems are being uploaded, and what OpenMHz endpoints they are servicing.
```
(info)   Setting up plugin -  Name: Openmhz	 Library file: libopenmhz_uploader.so
(info)   	[OpenMHz]	OpenMHz Server: https://api.openmhz.com
(info)   	[OpenMHz]	Uploading calls for: clmrn_I	 OpenMHz System: clmrn	 API Key: *****999
(info)   	[OpenMHz]	Uploading calls for: clmrn_G	 OpenMHz System: clmrn	 API Key: *****999
```


robotastic:
This is a great addition!

PR #910, opened 2024-01-24
Bcfy pluginfix

taclane:
The 5.0 curl upgrade is inadvertently using `api_key` instead of `apiKey` in the Broadcastify Plugin.  It also lacks the `metadata` upload, and attempts to upload the call audio prior to receiving the upload URL back from the server.

This reverts to the correct API key name, uploads the t-r metadata as before, and removes an early upload attempt for call audio.

Startup message have also been adjusted for clarity and identifying serviced endpoints:
```
(info)   Setting up plugin -  Name: broadcastify_uploader     Library file: libbroadcastify_uploader.so
(info)       [Broadcastify]    Broadcastify Server: https://api.broadcastify.com/call-upload
(info)       [Broadcastify]    Uploading calls for: clmrn_I     Broadcastify System: 12347568     API Key: ******b9
(info)       [Broadcastify]    Uploading calls for: clmrn_G     Broadcastify System: 99999999     API Key: ******b9
```

An additional fix was included to use a less-messy regex in the other two uploaders.

### DIFF
--- a/plugins/broadcastify_uploader/broadcastify_uploader.cc
+++ b/plugins/broadcastify_uploader/broadcastify_uploader.cc
@@ -311,6 +311,7 @@ public:
   }
 
   int parse_config(json config_data) {
+    std::string log_prefix = "\t[Broadcastify]\t";
 
     // Tests to see if the uploadServer value exists in the config file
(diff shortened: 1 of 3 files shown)


robotastic:
Nice fixes!! Thanks

PR #916, opened 2024-02-04
broadcastify_uploader.cc - minor log level adjustment

decoy4fun:
"Rejected" status currently logs at an 'info' level. This quick change will now show as an error if Broadcastify rejects your upload.


robotastic:
Agreed - that should probably be an Error

PR #922, opened 2024-02-12
Update GitHub workflow to build docker images on RC commit

ge0metrix:
Had to re-file this PR because workflows are run based on the workflow definition in the branch, not the main branch.


robotastic:
Let's give it a whirl!

PR #925, opened 2024-02-16
Plugin auto-build method for Trunk Recorder

taclane:
Trunk Recorder has been offering the ability for users to create and install their own plugins since early 4.x.  While the process is not difficult, and ensures that plugin code is unlikely to interfere with Trunk Recorder, there are some concerns where installed libraries and plugins may inadvertently fall out-of-sync with each other and lead to unexplained errors or crashes.

While it can be much easier to create plugins from within the Trunk Recorder source directory, this method often requires modifying elements of the upstream code to support finding and compiling these plugins.

This PR aims to strike a compromise between this and out-of-tree development by proposing a `/user_plugins` directory.  Upon `cmake`, this directory will be searched for any pre-staged plugin code and attempt to add it to the build and install process.  This will ensure such plugins keep current with any changes to the Trunk Recorder source, and help automate deployment (e.g. docker) by compiling user plugins and Trunk Recorder in the same step.

Compatible plugins are added to the directory via `git clone` or `git submodule add`, and as long as their dependencies are met, they will be compiled and installed along with the Trunk Recorder's built-in plugins.  Current user plugins may require minor changes to support building from `/user_plugins`, but it should be limited to:
- Adjusting plugin `#include`s or cmake `include_directories( ... )` to reference local headers
- Editing plugin `CMakeLists.txt` to use libraries created during the build process, not those installed to the system

This PR should also ease future development, as built-in plugins can simply be copied from `/plugins` to `/user_plugins` as a template and modified to suit.

### DIFF
--- a/docs/Plugins.md
+++ b/docs/Plugins.md
@@ -21,12 +21,12 @@ The [Built-in Plugins](#built-in-plugins) are compiled and installed when you se
 
 #### Plugin Object
 
-| Key     | Required | Default Value | Type                 | Description                                                  |
-| ------- | :------: | ------------- | -------------------- | ------------------------------------------------------------ |
-| library |    ✓     |               | string               | the name of the library that contains the plugin.            |
+| Key     | Required | Default Value | Type                 | Description                                                                                                                 |
+| ------- | :------: | ------------- | -------------------- | --------------------------------------------------------------------------------------------------------------------------- |
+| library |    ✓     |               | string               | the name of the library that contains the plugin.                                                                           |
 | name    |    ✓     |               | string               | the name of the plugin. This name is used to find the `<name>_plugin_new` method that creates a new instance of the plugin. |
-| enabled |          | true          | **true** / **false** | control whether a configured plugin is enabled or disabled   |
-|         |          |               |                      | *Additional elements can be added, they will be passed into the `parse_config` method of the plugin.* |
+| enabled |          | true          | **true** / **false** | control whether a configured plugin is enabled or disabled                                                                  |
+|         |          |               |                      | *Additional elements can be added, they will be passed into the `parse_config` method of the plugin.*                       |
 
 
 ## Built-in Plugins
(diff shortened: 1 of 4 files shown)


robotastic:
This looks pretty good! I should probably move some of the built-in plugins to be user-plugins, it would keep compile time down. I have to go look into git submodules now!

docs/Plugins.md
```diff
@@ -185,11 +185,59 @@ The matching simplestream config to send audio from talkgroup 58918 to TCP port
         }
 ```
 
+## Community Plugins
+Community plugins can extend the features of Trunk Recorder and allow customized workflows or analysis.  
+> As new plugins are developed, authors are encouraged to add to the below tables by submitting a PR to this document.
```
lilhoser, line 190: This creates an unfortunate dependency to PR against trunk/main. Maybe we could automatically pull in any README.md or just say "refer to user plugin documentation" in this section.

docs/Plugins.md
```diff
@@ -185,11 +185,59 @@ The matching simplestream config to send audio from talkgroup 58918 to TCP port
...
+
+#### External Plugins
+Plugins that are built out-of-tree and installed seperately from Trunk Recorder:
```
lilhoser, line 193: Are these technically "plugins"? Seems like they're just external applications that consume output.

docs/Plugins.md
```diff
@@ -21,12 +21,12 @@ The [Built-in Plugins](#built-in-plugins) are compiled and installed when you se
...
+| library |    ✓     |               | string               | the name of the library that contains the plugin.                                                                           |
 | name    |    ✓     |               | string               | the name of the plugin. This name is used to find the `<name>_plugin_new` method that creates a new instance of the plugin. |
-| enabled |          | true          | **true** / **false** | control whether a configured plugin is enabled or disabled   |
-|         |          |               |                      | *Additional elements can be added, they will be passed into the `parse_config` method of the plugin.* |
+| enabled |          | true          | **true** / **false** | control whether a configured plugin is enabled or disabled                                                                  |
+|         |          |               |                      | *Additional elements can be added, they will be passed into the `parse_config` method of the plugin.*                       |
 
 
 ## Built-in Plugins
```
lilhoser, line 32: What are the distinguishing characteristics of a "built-in plugin" and a "user plugin"? Is there a promotion process?

lilhoser: I would suggest filing tickets to move any plugin that does not meet the criteria for "built-in" to be a "user" plugin.

taclane: I am using the language to be consistent with existing references in the code or repository to describe those plugins normally included in an installation. 

But at present, the only distinguishing characteristic of a "built-in" plugin is that it is shipped with trunk recorder for the benefit of the majority of users, and can only be maintained by filing a PR against the repo. 

As fas as current candidates to move, the the `streamer` plugin has likely been broken since at least 4.6 without anyone noticing, and `stat_socket` can probably benefit by bundling it into https://github.com/TrunkRecorder/trunk-recorder-status-server, instead of splitting the functionality in two halves.

docs/Plugins.md
taclane: Yes.  There are a few examples such at my [MQTT fork](https://github.com/taclane/trunk-recorder-mqtt-status), that have a separate repository and are built against the headers and libraries installed with trunk-recorder.  Such plugins are functionally the same as the "built-in" plugins, and are loaded as libraries in the config.json.

I believe this was the intended method for plugin development, as it leaves the source entirely tree alone.  But it's got it's own set of issues, most notably falling out-of-sync with any trunk-recorder updates, and it forces the user to find and recompile any plugins whenever there are any changes to main libraries.

docs/Plugins.md
taclane: Probably the easiest way is to just link a trunk-recorder wiki page and let people add things themselves.

lilhoser:
LGTM.

robotastic:
LETS DO THIS! I will redo my plugins as User Plugins to test. For the v5.0 I will leave all the builtin plugins as is... but I will start working on moving most of them out to be User Plugins.

PR #930, opened 2024-02-21
Increment conventional call numbers on restart_call()

taclane:
Conventional calls are assigned a call number when they are first created, and continue to reuse that same `call_num` for every subsequent call.

While this reuse has no negative technical effects, it is not ideal from the standpoint of logging or reporting to have non-unique numbers as a call identifier.

This patch simply increments the `call_num` on each `Call_conventional::restart_call()`.


robotastic:
Brilliant!

PR #935, opened 2024-03-11
Save recorder and source number in Call_Data_t

taclane:
To aid in the analysis of `freq_error`, `signal`, and `noise` values in 5.0, it would be beneficial to track which SDR a call was recorded on.  This zero-indexed value will need to be compared against the source order in the config.json, but it will allow users interested in monitoring such information to correlate these metrics against unique devices.

This PR captures a call's source and recorder number in the `Call_Data_t` struct, and provides it to the call .json or any plugin seeking to use it.


robotastic:
This is a good add! I curious to see what kind of analysis people do with it

PR #937, opened 2024-03-17
Plugin smart retry

taclane:
When a plugin returns a failure for a `call_end` API function, trunk-recorder puts the call into a retry queue to attempt delivery two more times before abandoning the call and cleaning up.  However, this process does not note *which* plugins report a failure, and *all* upload plugins will be retried on each attempt until *all* plugins are successful.

Problems are often seen when a config uses multiple plugins (OpenMHz, broadcastify, rdio-scanner, mqtt, etc...).  Should any one of those endpoints go down (or have a problem with their SSL certs), all the other plugins will upload duplicate calls on each re-attempt.

This plugin records which plugins report a `call_end` failure, and on retry only attempts those to avoid sending duplicates to other services.


robotastic:
This is a great idea!

PR #939, opened 2024-03-21
don't remove call when upload failed

rosecitytransit:


taclane:
I don't know if there's an easy way to accomplish this without creating a new `archiveFailedUploads` config option, but just leaving the files in place might cause other issues.

The concern that comes to mind is that the default `tempDir` attempts to use a RAM-based filesystem.  An extended network outage is likely to break trunk-recorder, or degrade the whole system, if it isn't cleaning up after calls.

rosecitytransit:
That is a good point. Certainly an ongoing network issue (could be a configuration issue or OpenMHz/Broacastify issue too) could fill things up and cause havoc. For those who don't really have any local storage there may not be a place to put files that are in limbo. But it just seems bad to immediately, permanently delete things that haven't been successfully processed.

FWIW, the call that I was after was a 4015-second long open mic transmission that I had heard on my personal scanner but OpenMHz rejected it for being too large. One idea to handle that issue would be to have the `maxDuration` parameter apply to trunked as well as to conventional calls.

robotastic:
This is probably a good idea... especially after OpenMHz had all the troubles this summer. You don't want to be deleting local copies when the online service fails

PR #940, opened 2024-03-29
Potential docker enhancements & two-stage build

jquagga:
Good day,

The attached PR makes two slight changes only to the docker workflow.  The first is changing from privileged to device with rwm.  The latter is common in the sdr docker world and reduces the privilege of the container.  It is still running as root, but now as least it has the basic limitations. This change works both on the existing image (the one you would pull from docker pull today) and the two-stage build included (since ideally that build works identically, just smaller).

The second change is a conversion from a single to dual stage build.  This allows a ~2.5 GB reduction in size of the image.  This is accomplished by building trunk-recorder (and first gr-osmosdr) and copying those over to the final image.  The other change is that the final image only has the runtime packages and not the full dev packages.  One of the larger changes here is only adding the gnuradio dependencies that trunk-recorder actually needs instead of gnuradio.  gnuradio has a qt dependency if that gets added to the build you pull in the xorg stack and several kitchen sinks.  

It is possible when the ubuntu image moves forward, the associated apt-get library lines will need updating as there isn't a metapackage.  For example, Debian is on libgnuradio-network3.10.5 instead of .1, so that would need tweaking when the bump to the next ubuntu LTS happens.

This needs testing of course, but it's working fine for me so far and it should work the same as the previous image.  Just much smaller.

```
❯ docker images
REPOSITORY                  TAG       IMAGE ID       CREATED          SIZE
jquagga/trunk-recorder      4.7.1     ff7f233d4460   13 minutes ago   603MB
robotastic/trunk-recorder   latest    a87c76a618d6   3 months ago     2.96GB
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -74,5 +74,4 @@ WORKDIR /app
 ENV HOME=/tmp
 
 #USER nobody
-#CMD trunk-recorder --config=/app/config.json
-ENTRYPOINT ["/bin/bash"]
+CMD trunk-recorder --config=/app/config.json
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,23 +1,16 @@
-FROM ubuntu:22.04 AS base
+FROM ubuntu:22.04 AS builder
 
 # Install docker for passing the socket to allow for intercontainer exec
 RUN apt-get update && \
   apt-get -y upgrade &&\
   export DEBIAN_FRONTEND=noninteractive && \
-  apt-get install -y \
-    apt-transport-https \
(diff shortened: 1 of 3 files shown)


robotastic:
This is really cool! I didn't realize that those GR  library packages existed. @EricTendian would this break any of the things you are using the Docker image for?

EricTendian:
Agreed, this is awesome. @robotastic at this point I'm using the upstream Dockerfile, so this should be totally fine.

robotastic:
Let's give it a try! Report back if anyone has any issues.

PR #944, opened 2024-04-04
rc/v5.0 formatting fixes

taclane:
The `log_header()` formatter currently ends with a cyan ANSI color code, and most uses of the returned string are being double-tabbed in the console output. 

This PR removes the stray cyan code, as well as extra `/t` or whitespace in the boost logging messages.  Additional whitespace adjustments after a formatted header should not be necessary under most circumstances, as it already terminates with a tab in the resulting string.

```
BOOST_LOG_TRIVIAL(info) << loghdr << "Example status message.";
```

The openmhz and broadcastify plugins have been also reverted to utilize the pre-formatted `call_info.talkgroup_display` string that displayed correctly in the logs prior to the update.


robotastic:
🧢 I tip my hat good sir! Thank you for patching this up!

PR #945, opened 2024-04-09
(Very) minor update config.cc

penguinzephyr:
Added missing : and space for readability in logs.


robotastic:
There is no such thing as a minor PR - Thanks the fixes!!

Issue #950, opened 2024-04-22
Seemingly random core dumps.


penguinzephyr:
As per the title, TR seems to randomly crash out. I rebuilt TR from the 5.0 branch with debug symbols turned on and captured the below. Have attached the output of bt full.

[bt_full.txt](https://github.com/robotastic/trunk-recorder/files/15056365/bt_full.txt)

System is Ubuntu 22.04.4 LTS running as a VM with a single hackRF passed in as its source.

Happens indeterminately - this time was after 820 transmissions, earlier i got to ~90 before it crashed.

The TR error log doesnt display anything wrong - Simply restarts and throws the initialization sequence logs.

taclane:
I've been trying to track this down too.  Proximate cause appears to be:
`lib/op25_repeater/lib/p25p1_fdma.cc`
```cpp
                if ((sap == 61) && ((fmt == 0x17) || (fmt == 0x15))) { // Multi Block Trunking messages
                    if (blks > deinterleave_buf.size()
                        return; // insufficient blocks available

                    uint32_t crc1 = crc32(deinterleave_buf[1].data(), ((blks * 12) - 4) * 8); 
```
The attempt to access `deinterleave_buf[1]` if the zero-indexed array only has one item is throwing the assert error.  The core dumps I saw were identical to this.

An edit to the effect of for that second line might prevent it in the future.
```cpp
if ((blks > deinterleave_buf.size()) || (deinterleave_buf.size() == 1))
```

robotastic:
Nice work!! I will work on getting this upstreamed into OP25 too

taclane:
Reception on one of my control channels has been flakier than usual, so my going theory that it is probably just a receive issue and fragments of the MBT occasionally get lost.

But the assertion error does seem to be caused by an assumption that when a MBT header block is validated, there must _always_ be additional data in the `deinterleave_buf` to decode.  

If it weren't throwing this assertion error, it would probably end up ignoring the MBT as it very unlikely that the random data it would find in `deinterleave_buf[1]` would pass the crc checks anyway.

penguinzephyr:
Confirming the fixes applied in #951 appear to have fixed this issue and it appears to now be stable.Thanks!

PR #951, opened 2024-04-22
assertion error fixes

taclane:
1. Minor edit to `p25p1_fdma.cc` to address an issue seen in #950 where a MBT header may have been received without any following blocks.  Previously, this was handled when non-existent data in `deinterleave_buf[1]` did not pass the CRC check, but accessing the second item in a one-item array will trigger an assertion error in the development build.

2. Revert a previous edit in `p25_parser.cc` for `tsbk03` "Explicit Grant Update".  As seen in the [boatbod OP25](https://github.com/boatbod/op25/blob/88498ccf54ff6a3b4885a3420ae58ac8172b3b45/op25/gr-op25_repeater/apps/tk_p25.py#L721) fork, `ch1` should be used.  `ch2` will return a slot id of `-1`, and cause an assertion error when [p25p2_tdma.cc](https://github.com/robotastic/trunk-recorder/blob/346dcff4bf7bc82c7bab3fdbc6c012dce67a04b6/lib/op25_repeater/lib/p25p2_tdma.cc#L133-L137) encounters it later.

3. Also noted in the core dump for the above item, `patch_data` is included in all messages passed by the decoder, but only contains initialized variables when it encounters a patch-related opcode. As those opcodes may not be very frequent, the TrunkMessage.PatchData struct will generally carry random data as the message passes through trunk-recorder.  e.g.:
`patch_data = {sg = 140720929611216, ga1 = 140720929611056, ga2 = 97692074541337, ga3 = 97692100480080}`

### DIFF
--- a/lib/op25_repeater/lib/p25p1_fdma.cc
+++ b/lib/op25_repeater/lib/p25p1_fdma.cc
@@ -584,7 +584,7 @@ namespace gr {
                 blks = deinterleave_buf[0][6] & 0x7f;
 
                 if ((sap == 61) && ((fmt == 0x17) || (fmt == 0x15))) { // Multi Block Trunking messages
-                    if (blks > deinterleave_buf.size())
+                    if ((blks > deinterleave_buf.size()) || (deinterleave_buf.size() == 1))
                         return; // insufficient blocks available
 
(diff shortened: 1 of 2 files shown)


tadscottsmith:
In this particular packet, **ch2** refers to the transmit channel frequency, so it does make sense to use **ch1** (receive channel frequency).

PR #953, opened 2024-04-26
For testing / future - Ubuntu 24.04 Dockerfile

jquagga:
So when I proposed the dual-stage build for the Dockerfile I said that the libraries would probably change when Ubuntu rolled forward and Ubuntu just rolled forward.  I chased down the dependencies and this Dockerfile builds trunk-recorder and it seems to work in my light testing.  

This does change a bit about the prior Dockerfile.  This version doesn't build rtl-sdr or gr-osmosdr; it sources them from Ubuntu upstream.  Nobel is new enough to have rtlv4 support in it. That coupled with the fork of gr-osmosdr used in the previous Dockerfile not being happy building in 24.04 at the moment led me to move both back to upstream.


jquagga:
Ok it took a minute to get the dependencies right. It seems picky on 32-bit arm whereas amd64 worked with all 3 versions here. But it finally builds in workflow.

robotastic:
Awesome! Sorry I should have merged this in sooner

ekainz:
Did this intentionally remove support for SDRPlay in the Docker container?

Fails with `Failed parsing Config: SoapySDR::Device::make() no match`

PR #954, opened 2024-04-30
Control Channel Retune Handling

tadscottsmith:
This should help address the issue of the control channel trying to retune to an active traffic channel.

- Adds an INVALID_CC_MESSAGE message type for messages that are not correctly parsed and does not count them for control channel msgs/sec. Note that UNKNOWN messages are still counted in cases of correctly parsed messages with an unknown opcode.

- Adds a TDULC message type when the parser receives a TDU with Link Control on a channel it believes should be a control channel. The specs indicate there should not be Link Control messages on an active control channel, so this immediately causes a retune to the next available control channel.

### DIFF
--- a/trunk-recorder/main.cc
+++ b/trunk-recorder/main.cc
@@ -727,6 +727,16 @@ void handle_message(std::vector<TrunkMessage> messages, System *sys) {
       unit_answer_request(sys, message.source, message.talkgroup);
       break;
 
+    case TDU:
+    {
+      //Control Channel may have retuned to an active data channel. Do not count the message.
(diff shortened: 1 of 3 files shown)


robotastic:
Could you make this PR against the v5.0 branch instead? It looks like it is changing some of the functions that moved around  in the newer branch, making it trickier to merge in the future.

tadscottsmith:
Sure! I'll give that a shot tomorrow. I know the PR looks a lot nastier than it really is because I had to move the `retune_system` function definition above the `handle_message` function so I could reference it. It really only adds about 20 lines.

PR #956, opened 2024-05-06
Improve Control Channel Retune Handling

tadscottsmith:
Resubmitting this against v5.0

This should help address the issue of the control channel trying to retune to an active traffic channel.

- Adds an INVALID_CC_MESSAGE message type for messages that are not correctly parsed and does not count them for control channel msgs/sec. Note that UNKNOWN messages are still counted in cases of correctly parsed messages with an unknown opcode.

- Adds a TDULC message type when the parser receives a TDU with Link Control on a channel it believes should be a control channel. The specs indicate there should not be Link Control messages on an active control channel, so this immediately causes a retune to the next available control channel.

### DIFF
--- a/trunk-recorder/monitor_systems.cc
+++ b/trunk-recorder/monitor_systems.cc
@@ -546,7 +546,7 @@ void handle_call_update(TrunkMessage message, System *sys, std::vector<Call *> &
   }
 }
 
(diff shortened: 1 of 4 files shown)


robotastic:
Genius! This is a much better way to detect a control channel. Thanks!

PR #957, opened 2024-05-14
Broadcastify uploader fixes - 5.0

taclane:
Three relatively minor edits for the broadcastify plugin in rc/5.0

1. The Broadcastify plugin currently re-reads the call json from disk as part of the initial upload stage.  As of [#927], this disk read is no longer necessary and potentially undesirable now that the json can be directly dumped from from the `Call_Data_t` object into the curl object.  The dumped json is also stripped of whitespace and should be smaller than the formatted file Trunk Recorder generates for human review.
2. The `filename` field does not appear to be in use with the current Broadcastify API.  Additionally, `call_info.converted` provides an _absolute_ filename for the audio, and such uploads may unintentionally transmit details about the host system and/or home directories/usernames that are not required.  Broadcastify appears to ignore this field, and automatically generates filenames for audio as `[SystemId]/[startTime]-[talkgroup].m4a` using details provided during the upload.
3. The plugin does not presently use the `log_header()` formatting function to simplify logging statements.  These have been added where useful.

### DIFF
--- a/plugins/broadcastify_uploader/broadcastify_uploader.cc
+++ b/plugins/broadcastify_uploader/broadcastify_uploader.cc
@@ -143,10 +143,6 @@ public:
     curl_mime_name(part, "metadata");
 
     part = curl_mime_addpart(mime);
-    curl_mime_data(part, call_info.converted, CURL_ZERO_TERMINATED);
-    curl_mime_name(part, "filename");
-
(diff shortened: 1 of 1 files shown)


robotastic:
These are great changes!! thanks for contributing them

PR #966, opened 2024-08-04
Simplestream improvements

aaknitt:
- Clean up pass by value vs. reference to prevent data from changing during function execution
- Add call start and end events with additional metadata


robotastic:
Looks good!

PR #970, opened 2024-08-16
Patch handling improvement - override recordUnknown

aaknitt:
Override recordUnknown if an unknown talkgroup is the supergroup in a patch with a known talkgroup.


robotastic:
Nice add!

PR #973, opened 2024-09-13
Hostname resolution cache for upload plugins

taclane:
The curl library does not utilize the host's DNS cache, and relies on [an internal mechanism](https://curl.se/libcurl/c/CURLOPT_DNS_CACHE_TIMEOUT.html) to determine when to resolve a hostname query.

However, the curl _easy handles_ and _multi handles_ found in uploader plugins are not reused, nor do they exist long enough to benefit from this internal mechanism.  Every curl-based upload currently must start with a DNS lookup, no matter how recently the last one occurred.  Trunk Recorder users with a home DNS server likely noticed that the top domains may be `api.openmhz.com` or `api.broadcastify.com` by _**several orders of magnitude**_.

This patch implements a [shared curl cache](https://curl.se/libcurl/c/CURLSHOPT_SHARE.html) within each of the three main uploaders (OpenMHz, Broadcastify, Rdio Scanner).  These resolved hostnames will persist between uploads, and use a standard 300 second TTL to determine when to refresh the address.

The result of implementing this could reduce *millions* of unnecessary DNS lookups, and speed up uploads ~100 ms by maintaining a per-plugin hostname cache.

### DIFF
--- a/plugins/broadcastify_uploader/broadcastify_uploader.cc
+++ b/plugins/broadcastify_uploader/broadcastify_uploader.cc
@@ -21,10 +21,14 @@ struct Broadcastify_Uploader_Data {
   bool ssl_verify_disable;
 };
 
+boost::mutex curl_share_mutex;
+
 class Broadcastify_Uploader : public Plugin_Api {
   // float aggr_;
   // my_plugin_aggregator() : aggr_(0) {}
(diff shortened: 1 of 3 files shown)


robotastic:
This is great work!!

Issue #511, opened 2021-09-08
Make tau a configurable setting for Analog Recorders


robotastic:
Based on #501 it looks like the change to the correct tau value may have made audio to quiet by default. I will make this value configurable, and bring back the old default.

Issue #187, opened 2018-03-29
Analog VHF Low Band audio distorted


kg6uyz:
I am monitoring a couple analog VHF low band channels (42MHz) and have noticed the audio is distorted, some times the dispatcher is understandable but if they start speaking louder it distorts and the units in the field are always distorted, the low band channels are still 25khz wide band instead of 12.5khz narrowband like everything else now adays.  I am running a standard cheap RTL-SDR dongle with error correction based on GQRX that runs on my laptop.  I also noticed i have the same issue in GQRX with the audio, i can change the filter widths or the voice bandwidth and nothing seems to fix the distortion.  I rebooted to windows on that laptop and ran sdr-console, with filter changes i can get the audio to sound like normal radio traffic with no distortion.  I also noticed the PPM error correction is different between GQRX and SDR-Console, on GQRX i set it to 61 and on SDR-Console its -70.  Anybody else run into an issue like this?

robotastic:
Hmm... GQRX and Trunk Recorder are pretty similar on the backend. I would try to mess around with GQRX and see if there are setting that work well. I can try to make a recorder that works for low bands.

kg6uyz:
I played around with GQRX some more, the settings below work with no distortion;  

Input Controls: 
- LNA Gain: 40 dB

Reciever Options:
- Filter Width: Wide
- Filter Shape: Normal
- Mode: Narrow FM
 - - Max dev: 5KHz
 - - Tau: 530us
- AGC: off

robotastic:
Cool - I will go rummage through the GQRX code and see if I can make sense of it.

Dygear:
An LNA Gain of 40dBm seems super high to me. What antenna and SDR are you using?

I'm receiving audio in the VHF (150 - 150 MHz) range, and I don't have any distortion with a gain of 24dBm. As your getting analog channels, I'm just wondering if your clipping audio is due to you drowning the receiver's rf front end.

kg6uyz:
I receive 150 to 870mhz with no issues, i am running a 150 to 870mhz antenna, my low band (42mhz) dongle is a generic RTL2832u+R820t dongle.  It processes and gives me clean analog audio on sdr-console and gqrx with the same dongle and antenna setup.  Trunking-recorder gives me garbled audio.

jknepher:
Hi,  I noted this same problem and was able to resolve it.  Not sure of the right way to post back a patch, but have it ready to go if you are interested.
[...]

PR #975, opened 2024-09-16
Configurable de-emphasis time constant for analog recorders

algertc:
Related to #187 #511 

As mentioned in the above issues, many analog VHF low-band public safety frequencies use a higher de-emphasis time constant. After much confusion trying to set up a 42 MHz highway patrol stream and consistently getting distorted audio, I eventually found the value in the analog_reocorder set to 75µs. Adjusting it fixed the distortion, and the stream sounded as I had originally expected.

Looks like #511 was forgotten about, but it seems like a sensible addition as most broadcasters running into this problem are never going to find that file and rebuild. Would have saved me a boatload of time as a first-time user.

Example config.json:
```
"systems": [
        {
            "type": "conventional",
            "channels": [],
            "squelch": -60,
            "deemphasisTau": 0.000530
        }
]
```

### DIFF
--- a/docs/CONFIGURE.md
+++ b/docs/CONFIGURE.md
@@ -253,6 +253,7 @@ There is a list of available Plugins [here](./Plugins.md).
 | decodeFSync            |          | false                      | **true** / **false**                                                         | *Conventional systems only* enable the Fleet Sync signaling decoder. |
 | decodeStar             |          | false                      | **true** / **false**                                                         | *Conventional systems only* enable the Star signaling decoder. |
 | decodeTPS              |          | false                      | **true** / **false**                                                         | *Conventional systems only* enable the Motorola Tactical Public Safety (aka FDNY Fireground) signaling decoder. |
+| deemphasisTau              |          | 0.000750                      | number                                                        | *Conventional systems only* configure the de-emphasis time constant. 750µs for NFM (default), 75µs for WFM North America, 50µs for WFM most other regions.   |
 | enabled                |          | true                       | **true** / **false**                                                         | control whether a configured system is enabled or disabled                 |
 
 ***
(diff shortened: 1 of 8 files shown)


cschmittiey:
Hey I mentioned this in the discord as well but would you mind also documenting the configuration option in CONFIGURE.md? would be helpful for those of us that help out with configuration questions in the discord to be able to point to what the option does and when it's helpful for people to adjust.

taclane:
If you can, have you tried a value of
`"deemphasisTau": 0.000750`  (i.e. 750 μs)
in your patched config?  This is generally the "correct" value for [analog LMR systems](https://github.com/robotastic/trunk-recorder/issues/478).  

There was an attempt to correct it from the current `0.000075`, but everyone was used to the "brighter" and "louder" sound of 75 μs, so it got reverted.  It's possible that maybe it should be re-investigated anyway; lots of things have changed in the last few years.  

As far as documentation goes, these are generally not values a user is going to freehand.  The de-emphasis value on the receiving end is supposed to match the pre-emphasis value of the transmitter.  That way, it can correctly *unskew* the received audio to correct for how it was transmitted.  Some typical values for this include: 50 μs (WFM euro), 75 μs (WFM USA), 750 μs (NFM LMR)

Trunk Recorder using 75 μs is something of compromise.  It sounds reasonably okay, but it's probably not the ideal tau value for most systems.

algertc:
I have tried 750 and can't tell much of a difference. The reason I used 530 is because that was what GQRX suggested for narrow fm. I can't really find any good official source or spec for what is actually used.

Also, in terms of acclimation to the "brighter" sound, I'm not sure if that's for people trying to record broadcast or what, but my audio is still very good volume and not dull at all. Without the filter it's borderline unintelligible.

taclane:
Other thoughts:
- Trunk Recorder should default the analog time constant to 750 µs as originally discussed in #478.  It does not appear that the earlier issues concerning gain still exist, and it makes an immediate improvement in NFM audio quality for LMR systems with no additional configuration needed.
- Smartnet systems will be using 750 µs, so simply updating the default should have them covered.  Would be worthwhile to extend this configuration option to analog trunked recorders?

ZeroChaos-:
I think changing this to the spec default value of 750 and accepting this PR is the best of all choices.  Folks who want to tune it can, and we can use the spec default by...default.  Thanks for the work @algertc

PR #979, opened 2024-10-11
Update INSTALL-PI.md

Dygear:
Updated instructions for Raspberry Pi OS. Includes much more detail and added a section for UHD (Borrowed from the Arch (By The Way) instructions.).


robotastic:
Thanks for the upgrade - good point on the Ettus... with the Pi5 it really isn't crazy to add an Ettus to it

Dygear:
A Pi 5 and TR 5 are a match made in heaven. Listening to 4MHz (so more than a RTL-SDR can handle) and it’s rock solid.

PR #983, opened 2024-11-26
Fixed Raspberry Pi Docker Install Typo

avery1227:
### Summary
The documentation incorrectly listed ``--devices`` as an argument, which is not recognized by Docker. The correct argument is ``--device``, and this fix resolves that discrepancy.

### Why This Change Matters
The incorrect flag caused confusion and delays in usage, as ``--devices`` is not valid. By correcting this, we ensure the documentation is accurate and users can follow it without encountering errors.

### Personal Note
I spent more time than I’d like to admit figuring this out, so I hope this fix saves others from the same frustration.

### DIFF
--- a/docs/Install/INSTALL-DOCKER.md
+++ b/docs/Install/INSTALL-DOCKER.md
@@ -10,7 +10,7 @@ To get started, create a directory and place your **config.json** file there and
 
 ```bash
 docker run -it \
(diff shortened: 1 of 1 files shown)


robotastic:
Thanks for finding and fixing this... and sorry it me so long to merge

PR #984, opened 2024-11-26
auto-restart.sh - Enable Email Alerts

decoy4fun:
This adds crash email alert functionality to auto-restart.sh.  Some may find find alerts to be spammy which is why they are disabled by default. This not only alerts you to Trunk-Recorder having issues, but also locates and automatically attaches the latest log file (if available) so you can figure out what happened. 

Emails are disabled by default so if a user does not want emails, nothing changes. They can copy the file into their working directory normally and they are done. 

If they want email alerts every time Trunk-Recorder crashes/fails to restart, those can be enabled and configured as needed. 

Sleep time has been updated to 120 by default to reduce alerts/spam. 

Bulk of the work was done with the help of ChatGPT but I spent quite a bit of time fine tuning and adjusting things. This is something I wanted to share as I find it useful but no offense taken if we decide not to merge.

### DIFF
--- a/examples/auto-restart.sh
+++ b/examples/auto-restart.sh
@@ -1,6 +1,99 @@
 #!/bin/bash
 
-until ./trunk-recorder; do
-    echo "Server 'myserver' crashed with exit code $?.  Respawning.." >&2
-    sleep 5
+##############################CONFIGURATION##############################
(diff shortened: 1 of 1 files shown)


robotastic:
This is cool! thanks for sharing it

Issue #985, opened 2024-12-11
Duplicate code in Dockerfile?


mkmer:
Why do we do this 2x?  Can't we just copy?
https://github.com/robotastic/trunk-recorder/blob/a64b6392b7603c6d9c10f157293834687824cc0d/Dockerfile#L36-L65

robotastic:
Thanks for cleaning this up

PR #986, opened 2024-12-12
Docker Build: Only build librtlsdr 1x, install 2x

mkmer:
Improve Docker build - only build librtlsdr 1x, install 2x

Addresses issue: https://github.com/robotastic/trunk-recorder/issues/985


robotastic:
oh - good catch - I am not sure why it is getting built twice... your approach makes more sense. thanks

PR #992, opened 2025-01-06
Motorola opcode 0x05 0x09 fix

taclane:
TSBK opcode `0x05` is used by Motorola systems to broadcast channel status information.  This is incorrectly interpreted by Trunk Recorder as a "Unit To Unit Answer Request", and erroneously decodes and forwards the message as such.  As this Motorola status message occurs several times a second, it has a significant impact on any plugin or downstream attempt to analyze system metadata.

TSBK opcode `0x09` is presently unused by Trunk Recorder, but also has a frequent Motorola-specific variant that should be logged separately.  Known values are decoded as per Unitrunker display.

This PR breaks out these two Motorola opcodes to prevent them from being decoded as a standard P25 message.


robotastic:
oh interesting! this makes sense to me

PR #1001, opened 2025-02-23
Min Duration to Info instead of error

decoy4fun:
super minor update to log-level for one line. 

Brought this up on discord. I initially felt this should be a warning but others in discord stated the program is doing exactly what we told it to do and it should be info instead. Either way I feel like it shouldn't be an error.


robotastic:
This is a good compromise - def shouldn't be an error

Issue #1002, opened 2025-02-23
[BUG] - openssh-client package missing from docker container


jared-bloomer:
## Description
In the docker image `robotastic/trunk-recorder:5.0.1` the package `openssh-client` is missing. This creates a challenge when using the encode-upload.sh script because the `scp` command is not found. 

### Before installing `openshh-client` package
```
trunk-recorder  | [2025-02-23 17:22:08.026157] (info)   [hcps]  43C     TG:        200  Freq: 851.837500 MHz    C        Running upload script: /app/encode_upload.sh /app/media/hcps/2025/2/23/200-1740331326_851837500.0-call_43.wav /app/media/hcps/2025/2/23/200-1740331326_851837500.0-call_43.json /app/media/hcps/2025/2/23/200-1740331326_851837500.0-call_43.m4a
...
trunk-recorder  | Encoding: /app/media/hcps/2025/2/23/200-1740331326_851837500.0-call_43.wav
trunk-recorder  | Upload:
trunk-recorder  | /app/encode_upload.sh: line 12: scp: command not found
```

### After installing `openshh-client` package
```
trunk-recorder  | [2025-02-23 17:34:19.031367] (info)   [hcps]  385C    TG:        274  Freq: 852.050000 MHz    C        Running upload script: /app/encode_upload.sh /app/media/hcps/2025/2/23/274-1740332058_852050000.1-call_385.wav /app/media/hcps/2025/2/23/274-1740332058_852050000.1-call_385.json /app/media/hcps/2025/2/23/274-1740332058_852050000.1-call_385.m4a
trunk-recorder  | Encoding: /app/media/hcps/2025/2/23/274-1740332058_852050000.1-call_385.wav
trunk-recorder  | Upload:
```

### encode-upload.sh script that is being used
```
#! /bin/bash
echo "Encoding: $1"
filename="$1"
basename="${filename%.*}"
#mp3encoded="$basename.mp3"
mp4encoded="$basename.m4a"
json="$basename.json"
#eval "lame --preset voice" ${filename}
eval "nice -n 19 /home/luke/bin/ffmpeg -i $filename  -c:a libfdk_aac -b:a 32k -cutoff 18000 $mp4encoded > /dev/null 2>&1"

echo "Upload: $encoded"
eval "scp $json $mp4encoded username@server.com:~/smartnet-upload "
```

## Resolution

1.  Connect to the docker container with `docker compose exec -ti trunk-recorder bash`
2. Run the following commands
```
[...]

PR #1003, opened 2025-02-23
fix(add openssh-client and ffmpeg)

jared-bloomer:
## Description

* Adding openssh-client (Resolve [Issue #1002](https://github.com/robotastic/trunk-recorder/issues/1002))
* Adding ffmpeg (Installed at `/usr/bin/ffmpeg`)

### PR Checklist
* [x] Updated CHANGELOG.md
* [x] Tested Locally with Docker Image 5.0.1

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -34,7 +34,9 @@ RUN apt-get update && \
     libxtrx-dev \
     pkg-config \
     wget \
-    python3-six
+    python3-six \
+    openssh-client \
+    ffmpeg
 
 WORKDIR /src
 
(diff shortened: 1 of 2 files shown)


robotastic:
Makes sense to me!

Dockerfile
```diff
@@ -34,7 +34,9 @@ RUN apt-get update && \
     libxtrx-dev \
     pkg-config \
     wget \
-    python3-six
+    python3-six \
+    openssh-client \
+    ffmpeg
```
EricTendian, line 39: I think these packages need to go on line 53 of this file in the final Docker image, rather than this step which is just the intermediate build image.

EricTendian (commented):
@jared-bloomer I tried out this but it seems like the packages aren't included in the final image. I left a comment as to the cause.

PR #1013, opened 2025-03-26
static_cast start_time

taclane:
Explicit conversion of start_time to investigate error seen in buildx workflow.


taclane:
The builldx workflow was failing the arm7 compile earlier due to an implicit conversion could not be assumed on that architecture.

robotastic:
@taclane  is a hero! 🦸

PR #1020, opened 2025-05-21
Update INSTALL-PI.md - update keyring url

aaknitt:
https://discord.com/channels/928800455444791306/928807172807208962/1372617775209517086

Issue #1018, opened 2025-05-05
The Raspberry Pi Installation Instructions Need to be Updated


Dewey3:
While doing one of my 3 to 4 time a year fresh Raspberry Pi install routines, I noticed that the instructions for the `deb-multimedia` packages are now returning a "404 Not Found" error.  Going to the deb-multimedia website, I discovered that `deb-multimedia-keyring_2016.8.1_all.deb` in "Download the keys for the apt source and install them:" and "You can verify the package integrity with:" is now updated to `deb-multimedia-keyring_2024.9.1_all.deb`.  The install using the newer `deb-multimedia-keyring_2024.9.1_all.deb` completed without any errors, but I am not going to get a chance to see if it affects the running of TrunkRecorder until tomorrow or day after.

PR #1022, opened 2025-05-24
Updated multimedia keyring file.

Dygear:
This updates the deb-multimedia-keyring file needed for downloading the a few of the resources to convert wav files to mp3.

### DIFF
--- a/docs/Install/INSTALL-PI.md
+++ b/docs/Install/INSTALL-PI.md
@@ -66,13 +66,13 @@ deb https://www.deb-multimedia.org bookworm main non-free
 ```
 - Download the keys for the apt source and install them:
 ```bash
-wget https://www.deb-multimedia.org/pool/main/d/deb-multimedia-keyring/deb-multimedia-keyring_2016.8.1_all.deb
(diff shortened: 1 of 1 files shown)


Dygear:
Please, please, please. Double check before merging this with a few people going through the install steps to make sure this isn't doing anything stupid. This is updating a key ring file to accept the key provided by deb multimedia. It looks like to me like this is a benign change, but it does deal with platform trust so it should be looked at with extra care.

taclane:
> It looks like to me like this is a benign change, but it does deal with platform trust so it should be looked at with extra care.

The update does mirror the official instructions on https://www.deb-multimedia.org.  Pulling apart the deb package only shows the updated keyring files and the associated scripts to install it on the host machine.

This PR also overlaps with #1018 and #1020.

robotastic:
Thanks for pulling this together @Dygear  - I merged in #1019 since it also cleaned up the README

Dygear:
> Thanks for pulling this together @Dygear  - I merged in #1019 since it also cleaned up the README

Yeah that makes sense. My bad should have checked other PRs first.

PR #1026, opened 2025-06-24
Update Docs: multisite should be boolean, not string

ancker010:
Update CONFIGURE.md to properly reflect that the multisite parameter should be boolean, not a string.


robotastic:
great catch!!

PR #1027, opened 2025-07-06
Trivial edit to reflect default maxDev of 5000 for v5.0

gofaster:


robotastic:
Just to check - this looks like it is shifting the default from 4000 to 5000. I am reading that right?

gofaster:
It looks like default in trunk-recorder v5 is now 5000?  This PR just changes the logging and docs to match.
[https://github.com/TrunkRecorder/trunk-recorder/blob/8722ae5d64040a5754750c48dbc4d0d0559cae76/trunk-recorder/config.cc#L307](url)

robotastic:
You are totally right...
I got confusd becasue it included a change to config.cc but that it just the logging

PR #1030, opened 2025-07-23
Include frequency in CC decode rate warning

gofaster:
When logging Control Channel decode rate warnings, include the Control Channel's frequency.  This aids the identification of problematic frequencies.
i.e.  Monitoring multiple System objects, with the same shortName; each with multiple rotating control channels.

`[2025-07-23 10:13:45.080972] (error)   [VOGON]  freq: 8.548785e+08      Control Channel Message Decode Rate: 2/sec, count:  7`


taclane:
Great idea. That's an extraordinarily practical bit of diagnostic info even for single-site monitoring.

robotastic:
Love it! This is a great add!

PR #1037, opened 2025-08-21
Remove 'system' component for Boost 1.89

taclane:
With Boost 1.89 onward, the remaining `system` stub that `find_package()` previously found no longer exists and will cause `cmake` to return an error. This presently affects macOS (via homebrew), and other bleeding-edge distributions using new boost libraries.

The recommended resolution is to [remove `system` from the CMakeFile](https://github.com/boostorg/system/issues/132#issuecomment-3146378680), as it has not been necessary to include it as a required component since Boost 1.69.

Trunk Recorder already requires boost libraries much newer than 1.69, so it should not be needed to have specific logic in the CMakeFile to detect or configure older versions that it can no longer use.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -155,7 +155,6 @@ MESSAGE(STATUS "Configuring Boost C++ Libraries...")
 # Although not required on my system, some users have linking issues without
 SET(BOOST_REQUIRED_COMPONENTS
     thread
-    system
     program_options
     filesystem
     log_setup


robotastic:
Good catch - I am going to take a look too on the versions of Boost that are allowed in the CMake check. maybe we can also drop the older versions there

robotastic:
ah - so this is the line I was thinking of.... but  I don't really think we need it, or it may not be doing anything. The whole Cmakefile probably needs some love though, so I will leave this alone until then.

Let's just move this change in so things work again.

PR #1038, opened 2025-09-10
Updated changelog and version number

SimonSheehan:
I had seen some confusion in Discord about the version number, easy fix there.

Another user had some degree of confusion because of the changelog not being updated for a bit. I didn't see a full release notes for V5 anywhere so I cobbled something together that I hope may work, and pulled from the 5.0.1 and .0.2 notes. 

Hope this helps out a bit.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,45 @@
 Trunk Recorder ChangeLog
 ========================
 ### Unreleased
-* Added openssh-client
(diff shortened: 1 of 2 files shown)


robotastic:
Much appreciated! yea, I have changed to just using the change log on GitHub instead of maintaining this.... I should probably just remove this file or have a pointer to the Releases on Github.

PR #1039, opened 2025-09-12
simplestream updates to accomodate negative TGIDs 

aaknitt:
Negative TGIDs can sometimes be used on conventional systems to differentiate them from trunked TGIDs.


robotastic:
Good call!

PR #1042, opened 2025-10-03
Update Pi Instructions for Trixie

Dygear:
Includes the newer `.sources` file format, and for the moment a comment and possible deletion of the dkms-xtra section. Wasn't needed at least on the 500+, we shall see if it's needed on the 500/5/CM5.


Dygear:
Ok, these are good to go. Confirmed that the `deb-multimedia-keyring_2024.9.1_all.deb` is not needed as well as adding the `deb-multimedia.org` package repo. This is great as we are now using a stock install without any edits to repos for Trixie.

PR #1045, opened 2025-10-05
Include timestamp for start and stop of transmission in milliseconds.

TheGreatCodeholio:
I have added millisecond timestamps to for the start and the end of a transmission. Somewhat quick and dirty some optimizations could be added.

   "start_time": 1759627129,
    "stop_time": 1759627132,

Future changes would include deriving the start and stop time from the millisecond time, and changing the filenames instead of using a semi-hacky +1 second to avoid collisions. 

### DIFF
--- a/trunk-recorder/call_concluder/call_concluder.cc
+++ b/trunk-recorder/call_concluder/call_concluder.cc
@@ -70,6 +70,8 @@ int create_call_json(Call_Data_t& call_info) {
           {"phase2_tdma", int(call_info.phase2_tdma)},
           {"start_time", call_info.start_time},
           {"stop_time", call_info.stop_time},
+          {"start_time_ms", call_info.start_time_ms},
+          {"stop_time_ms", call_info.stop_time_ms},
           {"emergency", int(call_info.emergency)},
           {"priority", call_info.priority},
           {"mode", int(call_info.mode)},
@@ -402,10 +404,12 @@ Call_Data_t Call_Concluder::create_call_data(Call *call, System *sys, Config con
 
     if (it == call_info.transmission_list.begin()) {
       call_info.start_time = t.start_time;
+      call_info.start_time_ms = t.start_time_ms;
     }
 
     if (std::next(it) == call_info.transmission_list.end()) {
       call_info.stop_time = t.stop_time;
+      call_info.stop_time_ms = t.stop_time_ms;
     }
 
     Call_Source call_source = {t.source, t.start_time, total_length, false, "", tag};
--- a/trunk-recorder/call_impl.cc
+++ b/trunk-recorder/call_impl.cc
@@ -505,6 +505,8 @@ boost::property_tree::ptree Call_impl::get_stats() {
   call_node.put("duplex", this->get_duplex());
   call_node.put("startTime", this->get_start_time());
   call_node.put("stopTime", this->get_stop_time());
+  call_node.put("startTimeMs", this->start_time_ms);
+  call_node.put("stopTimeMs",  this->stop_time_ms);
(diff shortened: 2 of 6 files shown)


robotastic:
Clever! This makes sense to me. Let me go check how things look on the OpenMHz side. I am almost 100% sure it doesn't use the timestamp in the filename for anything, so the change in resolution will not have an impact.

TheGreatCodeholio:
Added some more work to this. 

Now we derive seconds from the new ms timestamp on a full call. 

filename is now `<seconds>.<ms>` format. i.e. `100-1759710634.379_154265000.0-call_24.xxx`

call metadata (json)
```
   "start_time": 1759710634,
    "stop_time": 1759710664,
    "start_time_ms": 1759710634379,
    "stop_time_ms": 1759710664885,
    "call_length": 31,
    "call_length_ms": 30506,
```

call_length is rounded, but call_length_ms 30.506 is a fully accurate playable length of the file. 

I figure with these changes, we have some time precision, but shouldn't effect anything down the pipeline too painfully as far as OpenMHZ or users custom scripts work.

Dygear:
I like this. I'll have to change some of my upload logic to the server, but I'm ok with that as it adds some extra details.

trunk-recorder/config.cc
```diff
@@ -345,8 +345,6 @@ bool load_config(string config_file, Config &config, gr::top_block_sptr &tb, std
         BOOST_LOG_TRIVIAL(info) << "Broadcastify API Key: " << system->get_bcfy_api_key();
         system->set_bcfy_system_id(element.value("broadcastifySystemId", 0));
         BOOST_LOG_TRIVIAL(info) << "Broadcastify Calls System ID: " << system->get_bcfy_system_id();
-        system->set_upload_script(element.value("uploadScript", ""));
```
Dygear: Why is the upload script dropped here?

Dygear:
I read some of the changes, and this is a massive patch. Some files delta like 6k lines of code. It also looks like it drops support for upload script for some reason? There is so much code in here, mostly because it updates a streamer plugin at the same time and calls into protobuf that I wouldn't feel comfortable reviewing. It feels like there are 3 or more patches in here than just adding millisecond timestamps. Is the protobuf streamer plugin updates required for the timestamp updates for some reasons or was this 3 patches squished into one PR?

taclane:
> It also looks like it drops support for upload script for some reason?

Since the upload script process is entirely fire-and-forget with no retry or followup, they've been working on a more extensible plugin solution: https://github.com/TheGreatCodeholio/tr_plugin_external_script

It's likely that some additional things have been inadvertently added to the PR branch since the initial push.  The [`backoff` fix](https://github.com/TrunkRecorder/trunk-recorder/pull/1045/commits/040066b47aac0396c54a4df7603d1229affdb7f9) should get applied though.  Retrying plugins at ~1 then ~2 minute intervals wasn't the intended ~2 then ~4 minute design, but it's been benign enough that it's generally gone unnoticed.

Dygear:
I like the millisecond changes, and I like that it's an extra field so current scripts still work. If that can be cherry picked out as the feature that is implemented in this PR that would be great. Removing something like Upload script probably requires more discussion for the moment. But, really, really great work @TheGreatCodeholio.

TheGreatCodeholio:
I am going to create a new pull request with just the ms_time feature added since I inadvertently added my work in progress here for modifying plugins and moving external scripts to a plugin. 

We will have to discuss further as a community what might be best for that, but in the end a plugin is the way to go. Maybe even a way to run them in a specific order and get the data from whatever ran last. This way transcript/storage/mqtt could be done locally on the fly and have additional keys added to JSON.

PR #1050, opened 2025-10-27
Colorized state messages and made output more consistent.

Dygear:
This colorizes the state message output that happens at a predefined interval to show how each source is being used. I also fixed some consistency issues with number width for when the count of recorders go over 10.


robotastic:
This is great!

PR #1055, opened 2025-11-06
P25 Autotune Improvements

taclane:
Several OP25 updates ago, the GNU Radio band-edge FLL filter was added to provide a limited amount of frequency correction for individual recorders.  At the completion of a call, Trunk Recorder would display the correction value, but do nothing further with the data.  These observations could be used to improve a source `error` value in the config file, but it required manual calculation and restarting the program.

This PR seeks to improve the process by collecting the last twenty correction values to build a running average for each source, and pre-load this correction to better center signals in advance of the FFL block.  This also leverages the existing t-r call flow and avoids trial-and-error approaches proposed in #217 or #271 to perform adjustments with minimal feedback.

While the FLL block is doing most of the work to facilitate this, maintaining a running average can be helpful with inexpensive SDRs that may exhibit a slow drift, or vary with the day/night ambient temperature cycle.  It also helps when setting up a new system, as it lessens the need to investigate a "good" initial `error` value.  Within a few minutes of active calls, it will also display a suggested `error` value to use next time you edit the config file.

At present, this only applies to P25 trunked/conventional/control channel recorders.  As the FLL data from other modes can be verified, this functionality can be expanded.  It is an opt-in feature, `"autoTune": true` is required in each source config to enable it.

While this has been reliably tested on fixed systems, a key component is that your transmission sources are accurate and stable enough to be used as references.  If transmission sources are inconsistent, it is possible that attempting to apply an average offset may negatively affect reception.


robotastic:
Huge addition - thanks @taclane !!

This also kills out one of the oldest PRs!

Dygear:
This is excellent! Time to get it up and running on my main boxes.

This going to get tagged out as 5.2? Autotune is a big deal should be worth a full point upgrade.

taclane:
> This going to get tagged out as 5.2? Autotune is a big deal should be worth a full point upgrade.

If we're thinking point releases, maybe we can hold off a couple days to roll in something for #939.  It's a little topical since sporadic network problems over the past weeks with broadcastify or openmhz may inadvertently be jamming up `tempDir` space when the default behavior is to leave behind failed uploads.

Dygear:
Yeah, that makes sense. Would be good to have these rolled into one release as they are both really good fixes / features.

PR #1056, opened 2025-11-06
System Information for P25 Parse Errors

taclane:
The debug message concerning P25 control channel parse errors can be vague if you are monitoring multiple systems.

While it may be possible to infer which system is experiencing an issue, this PR proposes to present the system name and control channel to remove any ambiguity.


robotastic:
Makes sense to me!

PR #1059, opened 2025-11-12
Formatter Updates

taclane:
Continuing the work of #1050:
- The two overloaded `format_state()` functions have been consolidated for ease of use and maintenance.
- Text colorization properly handles two-parameter `MONITORING` states.
- `formatter.h` now offers a selection of color constants (such as `Color::BLU`) for use in logging statements as an alternative to ANSI escape sequences.


robotastic:
Best use of an Enum ever...

Issue #1061, opened 2025-11-14
PR# 1060 seems lose JSON metadata for P25 Conventional.


gofaster:
For example, a channel that is defined  as `TG Number` 12 in a channelFile.

Pre PR#1060 JSON call metadata
```
{
  "freq": 454562500,
  "freq_error": -54,
  "signal": -18,
  "noise": -32,
  "source_num": 0,
  "recorder_num": 8,
  "tdma_slot": 0,
  "phase2_tdma": 0,
  "start_time": 1763043320,
  "stop_time": 1763043321,
  "emergency": 0,
  "priority": 0,
  "mode": 0,
  "duplex": 0,
  "encrypted": 0,
  "call_length": 1,
  "talkgroup": 12,
  "talkgroup_tag": "City FD 1",
  "talkgroup_description": "",
  "talkgroup_group_tag": "Fire Dispatch",
  "talkgroup_group": "City Fire",
  "audio_type": "digital",
  "short_name": "CITY",
  "freqList": [
    {
      "freq": 454562500,
      "time": 1763043320,
      "pos": 0.0,
      "len": 1.26,
      "error_count": 15,
      "spike_count": 0
    }
  ],
  "srcList": [
    {
      "src": 443082,
      "time": 1763043320,
      "pos": 0.0,
      "emergency": 0,
      "signal_system": "",
      "tag": ""
    }
  ]
}
```

Post PR#1060 JSON call metadata for the same system
```
{
  "freq": 454562500,
  "freq_error": -47,
  "signal": -19,
  "noise": -47,
  "source_num": 0,
  "recorder_num": 8,
  "tdma_slot": 0,
  "phase2_tdma": 0,
  "start_time": 1763077161,
  "stop_time": 1763077162,
  "emergency": 0,
  "priority": 0,
  "mode": 0,
  "duplex": 0,
  "encrypted": 0,
  "call_length": 1,
  "talkgroup": 1,
  "talkgroup_tag": "",
  "talkgroup_description": "",
  "talkgroup_group_tag": "",
  "talkgroup_group": "",
  "color_code": 0,
  "audio_type": "digital",
  "short_name": "CITY",
  "freqList": [
    {
      "freq": 454562500,
      "time": 1763077161,
      "pos": 0.0,
      "len": 0.9,
      "error_count": 0,
      "spike_count": 0
    }
  ],
  "srcList": [
    {
      "src": 449002,
      "time": 1763077161,
      "pos": 0.0,
[...]

PR #1062, opened 2025-11-14
Conventional P25 hotfix

taclane:
Recent DMR changes have allowed for dynamic conventional talkgroup assignments based on over-the-air group IDs.  While this may presently work for DMR, conventional p25 is not configured to manage this though the channelfile.csv as OTA talkgroups for simplex systems are often left as the default TG `1`.

This leaves the bulk of recent DMR changes undisturbed, and allows conventional p25 to return to the previous behavior where "virtual" talkgroups are arbitrarily assigned via channel frequency.  While it may be possible to use OTA p25 conventional talkgroups in some meaningful way in the future, t-r is not presently configured to do so.

Several log messages have also been demoted to the `debug` level to minimize clutter.

This should directly address #1061

### DIFF
--- a/trunk-recorder/gr_blocks/transmission_sink.cc
+++ b/trunk-recorder/gr_blocks/transmission_sink.cc
@@ -118,8 +118,8 @@ bool transmission_sink::start_recording(Call *call) {
   d_current_call_num = call->get_call_num();
   d_current_call_freq = call->get_freq();
   d_conventional = call->is_conventional();
-  if (d_conventional && ( (call->get_system_type() == "conventionalP25") || (call->get_system_type() == "conventionalDMR") )) {
-    BOOST_LOG_TRIVIAL(debug) << "transmission_sink::start_recording - Conventional flag is set for a digital system type - dynamically assigning talkgroups";
+  if (d_conventional && (call->get_system_type() == "conventionalDMR")) {
+    BOOST_LOG_TRIVIAL(debug) << "transmission_sink::start_recording - Conventional DMR - dynamically assigning talkgroups";
     d_current_call_talkgroup = 0;
     d_current_call_talkgroup_display = "N/A";
     d_current_call_talkgroup_encoded = 0;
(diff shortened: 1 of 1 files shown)


ZeroChaos-:
Why intentionally override the metadata in the packets?  Even on conventional you can have multiple talk groups and certainly care about the difference between security and maintenance talk groups.

taclane:
> Why intentionally override the metadata in the packets? Even on conventional you can have multiple talk groups and certainly care about the difference between security and maintenance talk groups.

With conventional P25, the frequency typically is "the talkgroup".  Remapping p25 by channel is the only effective way to manage a system like:

| Freq. | Name | Type | TG |
| --- | --- | ---| ---|
|451.500 | PD Dispatch | P25 | 1 |
|452.000 | PD OPS | P25| 1 |
|452.250 | Fire | P25| 1 |
|453.375 | EMS | P25| 1 |

While it may be possible that the channelfile.csv system can be improved to better accommodate a typical situation like the above, there is no available workaround when all the channels in a conventional p25 system report back as the default TG 1.

robotastic:
Ah! I didn't realize that Conventional P25 didn't really make use of the talkgroup ID and went by frequency/channel instead. Thanks for fixing this.

taclane:
> Ah! I didn't realize that Conventional P25 didn't really make use of the talkgroup ID and went by frequency/channel instead. Thanks for fixing this.

It *can*, but often does not in fielded systems.  A huge chunk of conventional p25 are "digitized" analog systems where the thing that discriminates traffic is the frequency in use.

Some of this is even further codified in publications like the [NIFOG](https://www.cisa.gov/safecom/field-operations-guides), where conventional p25 channels must be left on TG 1.  There is no practical way to monitor something like that by TG alone when you have a few dozen frequencies all reporting as the default value.

Single channel p25 systems with multiple talkgroups do exist, but they wouldn't be well-served by the present channelfile.csv implementation.  We'd need a better hybrid approach to support that in future.

PR #1065, opened 2025-11-20
Change source log in simplestream to debug log

EricTendian:
When using the simplestream plugin, I noticed that my logs were being spammed with messages like "no source found for call - leaving src as -1" when analog transmissions were being streamed, and sometimes the other log for digital streams. These logs don't seem to be necessary outside of debugging so I'm changing the log level from info to debug.


Dygear:
100% agree with this PR.

robotastic:
Makes sense to me!!

PR #1066, opened 2025-11-21
Add ms time 2 (The Electric Boogaloo)

TheGreatCodeholio:
Lets try this again. These changes should ONLY be the ms_time portion of the previous update. This leave existing timestamps in the metadata json and adds new keys for MS timestamps. 

It also changes the file name when saved to include the timestamp in ms when the call started. This is due to a bit of a hacky way of adding +1 sec to avoid collisions for filenames. Now collisions should not happen. 

"start_time": 1759627129,
"stop_time": 1759627132,


gofaster:
> It also changes the file name when saved to include the timestamp in ms when the call started. This is due to a bit of a hacky way of adding +1 sec to avoid collisions for filenames. Now collisions should not happen.

Please show an example of the new file name format.  I have post-process scripts that extract date/time from the name.

taclane:
Might need to drop https://github.com/TrunkRecorder/trunk-recorder/pull/1066/commits/06c398658aa0adbbc945e58bb5142ee05566dfcf; that's bringing a lot of the unintentional changes from the first PR.

PR #1067, opened 2025-11-21
Millisecond time and backoff fix, the third time is the charm? 

TheGreatCodeholio:
Lets try this yet again. These changes should ONLY be the ms_time portion of the previous pull request. This leaves existing timestamps in the metadata json and adds new keys for MS timestamps.

It also changes the file name when saved to include the timestamp in ms when the call started. This is due to a bit of a hacky way of adding +1 sec to avoid collisions for filenames. Now collisions should not happen.

"start_time": 1759627129,
"stop_time": 1759627132,

file name format: we add a .xxx after the initial timestamp
`300-1763759979.250_155415000.0-call_6.wav`

Exponential backoff fixed using XOR instead of wrong symbol. This should now doa  proper back off and retry on failed call workers after call is finished. 

`long backoff = ((1 << call_info.retry_attempt) * 60) + jitter;`


Dygear:
This all looks good to me. Nice clean PR. Great work @TheGreatCodeholio

robotastic:
Code was merged in via #1093

PR #1069, opened 2025-11-27
Fix gain precision

taclane:
Trunk Recorder currently reads gain values from the config as an integer, ignoring any decimal component of the number.  The entered value is truncated (rounded down), and passed on to the underlying driver to determine what available setting to use. 

While the net effect of truncation may be minimal (sometimes even rounding _down_, then rounding _up_ by the driver), it leads to some ambiguity between the config value, and what is used by the device.  In other instances, the valid gain setting of `0.9` for RTL devices will get rounded down to `0` (also a valid setting), and stay that way.  

While this change applies all device types, it is especially important to RTL devices where valid gain settings do not land on whole integers.

```
Gain Stage: LNA supported values: 0 0.9 1.4 2.7 3.7 7.7 8.7 12.5 14.4 15.7 16.6 19.7 20.7 22.9 25.4 28 29.7 32.8 33.8 36.4 37.2 38.6 40.2 42.1 43.4 43.9 44.5 48 49.6
```

Whether or not driver-valid settings are integers, the underlying osmosdr, UHD, and soapy libraries already use double precision values to manage gain.  This PR just ensures that any value with a decimal component is presented to the driver as entered.

Additionally, t-r will read back the set gain value for osmosdr devices and note when it differs from the config value if the driver performed an adjustment.

### DIFF
--- a/trunk-recorder/source.cc
+++ b/trunk-recorder/source.cc
@@ -320,7 +320,7 @@ void Source::set_gain(double r) {
     cast_to_osmo_sptr(source_block)->set_gain(gain);
     double current_gain = cast_to_osmo_sptr(source_block)->get_gain();
     if (current_gain != gain) {
-      BOOST_LOG_TRIVIAL(error) << "Requested Gain of " << gain << " not supported, closest value is: " << current_gain;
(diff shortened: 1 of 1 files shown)


robotastic:
great catch!

PR #1070, opened 2025-11-30
Do not retry OpenMHz uploads for configuration errors

taclane:
This addresses a bug where OpenMHz configuration errors (invalid API keys, incorrect system IDs, bad filenames, ignored talkgroups) will return a call to the retry queue instead of skipping and moving on.

In the event of an upload failure, the plugin will parse the reply from the server and compare it against a list of select errors that will not require further handling.  All other errors (such as connection issues) will return the call to the retry queue.

This is a helpful change because the OpenMHz admin panel includes a configuration option for `Ignore Unknown Talkgroups`.  Should that option be enabled, Trunk Recorder presently retries a call from an unknown talkgroup until it reaches the maximum number of attempts.  The broadcastify plugin uses a similar form of error handling to determine that `SKIPPED` or `REJECTED` calls need not be reattempted.

Preventing these calls from being reattempted will reduce some load on the OpenMHz API, and ensure that temporary files on the local device are cleaned up without lingering another ~6 minutes until `MAX_RETRY`.

### DIFF
--- a/plugins/openmhz_uploader/openmhz_uploader.cc
+++ b/plugins/openmhz_uploader/openmhz_uploader.cc
@@ -305,6 +305,27 @@ public:
       }
     }
     std::string loghdr = log_header(call_info.short_name,call_info.call_num,call_info.talkgroup_display,call_info.freq);
+
+    // Configuration errors - skip the retry queue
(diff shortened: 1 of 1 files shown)


Dygear:
Clean, I like it.

robotastic:
great call!!

PR #1072, opened 2025-12-13
Standardize trunk-recorder.service & Contextualizes Comments & Commands.

Dygear:
1. Updates the `trunk-recorder.service` file to use the `make install` location of trunk-recorder.
 a. Also makes the user `trunk` and not `luke`, so it expects the user `trunk` to exist.
2. Added comments for troubleshooting core dumps and usage instructions.
3. Sorta reverts the old `auto-restart.sh` but renames it so people can get a dead simple script to start off from.

### DIFF
--- a/examples/auto-restart-simple.sh
+++ b/examples/auto-restart-simple.sh
@@ -1,0 +1,6 @@
+#!/bin/bash
+
+until ./trunk-recorder; do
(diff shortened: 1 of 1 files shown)


robotastic:
... yea - I probably shouldnt bake my name into the example ... 🙃

Dygear:
I mean, it's fair. That file was I think from 10 years ago.

PR #1073, opened 2025-12-13
Uses the "Not recording talkgroup - Priority is -1 (Disabled)." language

Dygear:
to make it more explicit.


Dygear:
Notifies the user that the priority that `-1` is the effective disabled flag for that talkgroup. This follows the documentation but puts it in front of the user as well.

From:
> "Not recording talkgroup. Priority is -1."

To:
> "Not recording talkgroup - Priority is -1 (Disabled)."

As a request was made in discord for this to make be more explicit in the logs.

robotastic:
This is clearer - thanks!

PR #1074, opened 2025-12-19
Motorola OTA Alias support for P25P1/P2

taclane:
This (finally) adds support for Motorola OTA aliases on P25 systems.  These aliases are broadcast on the voice channel and visible to units in the talkgroup.  If used in the system, they can supplement the current `unitTags` metadata about participants in a call.

To this effect, the following changes are proposed:
- P25 TDMA/FDMA decoders to parse relevant opcodes, and send data into a separate message queue for processing downstream
- A c++ implementation of @ilyacodes decoder algorithm, with a pre-computed inverse lookup table for added efficiency
- Configuration settings to retain aliases between sessions (`unitTagsOTA`), and toggle whether OTA or user-defined aliases take precedence (`unitTagsMode`)
- Rewrite of `unit_tags.cc` to use the improved csv handler and support the `unitTagsOTA` cache

General operation is as follows:
- TDMA/FDMA decoder detects an alias, it is turned into plain text and provided to the parent `shortName`
- If it is a new alias, it is added to an internal list, and appended to the `unitTagsOTA` file if configured
- Calls including that radio ID will have the alias text added to the call metadata
- On each restart, trunk-recorder will clean up the alias cache, sorting entries and trimming duplicates
- If a `unitTagsOTA` entry has incomplete metadata (i.e. started as a simple [unit, alias] list), records will be enriched with additional information as it is discovered

While this PR does not handle other types of aliases, enabling the message queue will allow further analysis and development of information present on the voice channel.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -259,6 +259,7 @@ list(APPEND trunk_recorder_sources
   trunk-recorder/talkgroups.cc
   trunk-recorder/unit_tag.cc
   trunk-recorder/unit_tags.cc
+  trunk-recorder/unit_tags_ota.cc
   trunk-recorder/plugin_manager/plugin_manager.cc
   trunk-recorder/call_concluder/call_concluder.cc
   trunk-recorder/autotune.cc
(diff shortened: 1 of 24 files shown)


robotastic:
🎅

Dygear:
Champagne for everyone!

PR #1077, opened 2026-01-02
get_unit_tags() methods for manual or OTA aliases

taclane:
As a followup to the OTA PR #1074, unit tags and OTA tags should provide a function to return the respective alias vectors if needed by another process.

While not presently used in trunk-recorder itself, this would allow plugins to iterate across
`sys->get_unit_tags()` or 
`sys->get_unit_tags_ota()`
much the same way it can presently do so with
`sys->get_talkgroups()`

A example use would be to periodically poll the length of OTA tags, then display or emit the *N* newest entries from the front of the vector if a change in length is noted.

### DIFF
--- a/trunk-recorder/systems/system.h
+++ b/trunk-recorder/systems/system.h
@@ -151,6 +151,8 @@ public:
   virtual std::vector<double> get_channels() = 0;
   virtual std::vector<double> get_control_channels() = 0;
   virtual std::vector<Talkgroup *> get_talkgroups() = 0;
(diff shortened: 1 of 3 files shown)


Dygear:
Very much +1 for this. I think I pinged @taclane about it right after the first beta went out on his repo. It would be great to have a way to get the OTA list from the plugin interface. Happy to see that land here.

Dygear:
@robotastic for what it's worth, I did read over the code and compiled it. This all looks good to me.

robotastic:
Good prod - let me go roll it in!

PR #1079, opened 2026-01-09
Update to support future Rdio alias uploads

taclane:
Trunk Recorder currently does not send alias data (OTA or manual) via the Rdio Scanner plugin. The Rdio server does not parse aliases via the API, but the intent is to correct this on both ends.

This is an implementation of a patch by @poisonednumber https://github.com/poisonednumber/trunk-recorder/commit/0e07906a051f88db3836b3484cde7fece6d55efb with additions to also cover a future "units" endpoint for this data as proposed in the unreleased [Rdio v7 API](https://github.com/chuot/rdio-scanner/blob/e0ff30a43c832f486e4aa1557d27c01352b7a92c/docs/api.md).

Enabling both "sources" and "units" does not seem advisable at this time.  Rdio 7v may process them both, leading to redundant metadata being added to the call.  A PR has been proposed for ThinLineRadio to test this feature while independent development continues with Rdio Scanner https://github.com/Thinline-Dynamic-Solutions/ThinLineRadio/pull/48.

### DIFF
--- a/plugins/rdioscanner_uploader/rdioscanner_uploader.cc
+++ b/plugins/rdioscanner_uploader/rdioscanner_uploader.cc
@@ -86,6 +86,11 @@ public:
     source_list << std::fixed << std::setprecision(2);
     source_list << "[";
 
+    std::string unit_list_string;
+    std::ostringstream unit_list;
+    unit_list << std::fixed << std::setprecision(2);
+    unit_list << "[";
+
     std::ostringstream patch_list;
     std::string patch_list_string;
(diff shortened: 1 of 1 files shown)


taclane:
Minor update, https://github.com/Thinline-Dynamic-Solutions/ThinLineRadio/pull/48 was approved, and this code correctly passes it alias data via "sources".

Including alias data will not affect unpatched rdio servers, as the parser will simply ignore extraneous `"tag"` fields.

PR #1081, opened 2026-01-14
Adding direct support for libiio based radios

ratmandu:
This adds support for libiio based radios, including but not limited to: ADALM Pluto, fmcomms2, LibreSDR, ZynqSDR, and many clones.
Directly using gnuradio-iio to interface with these radios allows use of multiple devices at once (soapy seems to only support a single device at a time) over USB or Ethernet connection (potentially local hardware as well, but that is untested as of now).

I've also added a bit of documentation for config, an example config, and added dependencies to the docker file. If the required dependencies for gnuradio-iio (or gr-iio itself) are not installed, cmake will build without iio support so it shouldn't break things for people updating T-R in the future.


robotastic:
Interesting! I have a Pluto - does this only work on Linux or would it work on a Mac too? I have a bit of travel coming up but I can test it out after that. It would be great if some other folks can too.

ratmandu:
It should work on Mac as well, with libiio installed and gnuradio compiled with iio support, but I have not tested it on Mac myself.

robotastic:
I failed miserably on getting it running on my MacBook - going to pivot to the Linux box.

ratmandu:
> I failed miserably on getting it running on my MacBook - going to pivot to the Linux box.

Yeah, I had issues as well, but it seems to mainly stem from getting gnuradio built with libiio/gr-libiio. I tried both homebrew and macports and had no luck, I'll take a look at those and see if I can figure out whats going on if I get some time

ancker010:
Just curious if there's been any movement on this.

robotastic:
Were doing this! I still haven't had a chance to test it with my Pluto...

PR #1082, opened 2026-01-15
Save Config Location on Startup

taclane:
After reading the config, trunk recorder does not retain the filename of the json loaded.  Should there be interest in an interactive config editor in the future, it will be important to know what file was used to perform read/write operations on it.

While there are ways to extract parameters used during process execution, it is far easier to just save the filename itself instead of maintaining various OS-specific workarounds.

The PR simply adds a new `config_file` string to the global Config struct, and populates it on startup.


robotastic:
Makes sense to me!

PR #1083, opened 2026-01-16
Add optional whitelist/blacklist for talkgroups on RDIO upload plugin…

TheGreatCodeholio:
This PR updates the Rdio Scanner uploader plugin with (1) per-system talkgroup
allow/deny filtering using glob-style patterns, and (2) targeted libcurl
robustness + observability fixes (timeouts, better error reporting, and more
correct status/result handling). Existing configs continue to work unchanged.

### Rdio Scanner Plugin (rdioscanner_uploader) — Changes in this PR
===============================================================================

1) Per-system talkgroup filtering (new allow/deny keys + glob patterns)
----------------------------------------------------------------------
Added two optional keys on each Rdio “system” object:

- talkgroupAllow (array)
  If provided and non-empty, the talkgroup MUST match at least one pattern
  to upload. If it doesn’t match, the call is skipped.

- talkgroupDeny (array)
  If provided and non-empty, any matching talkgroup is skipped.

Patterns are glob-style and compiled to fully-anchored regex (^...$):
- * matches any number of characters
- ? matches exactly one character

Filter evaluation order:
1) allow check (if present)
2) deny check (if present)

Additional notes:
- Matching is performed against the numeric talkgroup ID coerced to a string
  (e.g., 50712), so pattern lists can contain strings or numbers.
- When a call is skipped due to filters, we log an info message that includes
  the talkgroup ID for easy debugging.

2) libcurl robustness / observability / correctness improvements
---------------------------------------------------------------
The upload path now includes these changes:

Observability:
- Adds CURLOPT_ERRORBUFFER and includes curl_err=... in error logs when present.
- Error logs now include HTTP status code in the message context.

Thread-safety / stability:
- Sets CURLOPT_NOSIGNAL=1 to avoid signal-related issues in threaded usage.

Timeouts:
- Sets CURLOPT_CONNECTTIMEOUT_MS (e.g., 15000ms).
- Sets CURLOPT_TIMEOUT_MS (e.g., 120000ms).

Correctness fixes in multi + HTTP handling:
- Extracts the per-transfer CURLcode from curl_multi_info_read() so we can tell
  whether the transfer itself failed vs. a non-2xx HTTP response.
- Captures CURLINFO_RESPONSE_CODE BEFORE cleaning up the easy handle
  (the old flow attempted to read it after cleanup).

Success criteria expanded:
- Treats any successful HTTP status as success via is_success_http_status(...),
  and explicitly supports 202 Accepted (logs “Accepted (202) - stub cached” vs.
  “Success” for other success statuses).

Safer lifetime for mime string inputs:
- Uses explicit std::string variables for fields passed into curl_mime_data()
  (audioType/dateTime/system/talkgroup/etc.) so the backing memory is guaranteed
  to remain valid until after the request completes.

Net effect
----------
- Adds per-system talkgroup allow/deny filtering with glob patterns.
- Makes uploads more reliable (timeouts + NOSIGNAL), fixes status handling, and
  improves diagnostics (curl error buffer + clearer error/success logging).

### DIFF
--- a/docs/Plugins.md
+++ b/docs/Plugins.md
@@ -45,30 +45,34 @@ This plugin makes it easy to connect Trunk Recorder with [Rdio Scanner](https://
 
 *Rdio Scanner System Object:*
 
-| Key              | Required | Default Value | Type   | Description                                                                                                                                                                                                 |
-| ---------------- | :------: | ------------- | ------ | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| systemId         |    ✓     |               | number | System ID for Rdio Scanner.                                                                                                                                                                                 |
-| apiKey           |    ✓     |               | string | System-specific API key for uploading calls to Rdio Scanner. See the ApiKey section in the Rdio Scanner administrative dashboard for the value it should be.                                                |
-| shortName        |    ✓     |               | string | This should match the shortName of a system that is defined in the main section of the config file.                                                                                                         |
-| talkgroupWhitelist |        | []            | array  | Optional allow-list of talkgroups to upload for this system. If set (non-empty), the talkgroup **must** match at least one pattern to be uploaded. Patterns are glob-style (supports `*` and `?`).        |
-| talkgroupBlacklist |        | []            | array  | Optional deny-list of talkgroups to block upload for this system. If set (non-empty), any matching talkgroup will be skipped. Patterns are glob-style (supports `*` and `?`).                             |
+| Key             | Required | Default Value | Type  | Description                                                                                                                                                                                                 |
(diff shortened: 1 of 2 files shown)


Dygear:
Clean! Nice!

TheGreatCodeholio:
54% less racist 

now uses talkgroupAllow and talkgroupDeny for config names. 

updateed original request note with Ai Slop, because I didn't want to type out all the changes I made with curl.

robotastic:
@Dygear - were you able to give the code a run? @TheGreatCodeholio - Thanks for including changes to the docs too!!

Dygear:
> @Dygear - were you able to give the code a run? @TheGreatCodeholio - Thanks for including changes to the docs too!!

I don't use rdio, just read over the code. Very clinical in its changes / additions / edits. I like it.

PR #1084, opened 2026-01-16
Broadcastify allow/deny talkgroup filters (glob patterns)

TheGreatCodeholio:
This PR modernizes and extends Broadcastify talkgroup filtering by introducing
allow/deny terminology and glob-style pattern matching, while preserving
backward compatibility with existing configs.

What changed
- Added per-system talkgroup filters:
  - broadcastifyAllow: if non-empty, the talkgroup MUST match at least one
    pattern or the upload is skipped
  - broadcastifyDeny: if non-empty, any matching talkgroup is skipped
- Patterns are glob-style (compiled to anchored regex):
  - * matches any length
  - ? matches exactly one character
- Filter order:
  1) allow check (if present)
  2) deny check (if present)

Docs / examples
- Updated config docs to describe broadcastifyAllow / broadcastifyDeny as arrays
  of string/number.
- Added examples showing both * and ? usage (e.g., "507*" and "12?45").

Notes
- Filtering compares against the numeric talkgroup ID coerced to a string
  (e.g., 50712) so behavior is consistent whether config entries are strings
  or numbers.
- When a call is skipped due to filters, we log an info message including the
  talkgroup ID for easier debugging.

Requesting review on naming consistency and the compatibility-key priority
ordering, but functionally this is a safe, additive upgrade.

54% less racist then before.


robotastic:
This looks straightforward to me and a good idea. I don't have Broadcastify uploading set up on my machine. Can someone give this code a run and see if it also works for them?

PR #1085, opened 2026-01-17
OP25 based SmartNet parsing

robotastic:
updated to borrow from the OP25 SmartNet parser. The actual RF chain still doesn't match 1:1 - I had some issues with the demod. I am unable to get the OP25 FM demod approach working, so I have fallen back to using the current PLL demod. 
https://github.com/TrunkRecorder/trunk-recorder/tree/dev/smartnet-refactor

Here is the OP25 approach. 

Config:
https://github.com/boatbod/op25/blob/198c92379ddf5a07e60f90cfa9e965a052e61b30/op25/gr-op25_repeater/apps/smartnet_example.json#L8-L15

Filter:
https://github.com/boatbod/op25/blob/198c92379ddf5a07e60f90cfa9e965a052e61b30/op25/gr-op25_repeater/apps/p25_demodulator.py#L156-L165

Chain:
https://github.com/boatbod/op25/blob/198c92379ddf5a07e60f90cfa9e965a052e61b30/op25/gr-op25_repeater/apps/p25_demodulator.py#L502


robotastic:
Cancel that - I now have the FM Demod working great... doing that instead

PR #1088, opened 2026-01-22
Use plugin names when possible

taclane:
Plugins have a `name` field in the config that was initially used to create an instance of the plugin.  This no longer seems to be the case, so it can be repurposed to allow plugins to use it for identification and logging.

This PR will better allow plugins to utilize the `name` field in the config to set a friendly plugin name and display it later.  The primary benefit of this would be to better identify when multiple instances of the same plugin are sending data to different servers within the logs:
Screenshot 2026-01-22 at 11 06 59 AM

If left blank, the `name` will default to the base filename.  Individual plugins may choose to handle the default value by the exchanging it for a friendly display name instead.

This has been implemented for the three main upload plugins (Rdio, Broadcastify, OpenMHz).  Setting a custom name for the latter two plugins can be accomplished by manually creating a plugin entry such as:

```json
{
  "name": "BCFY",
  "library": "libbroadcastify_uploader.so",
  "broadcastifyCallsServer": "https://api.broadcastify.com/call-upload",
  "systems": [
    {
      "shortName": "mysystem",
      "broadcastifySystemId": 1123,
      "broadcastifyApiKey": "API-key"
    }
  ]
}
```
or
```json
{
  "name": "OpenMHz.com",
  "library": "libopenmhz_uploader.so",
  "uploadServer": "https://api.openmhz.com",
  "systems": [
    {
      "shortName": "mysystem",
      "openmhzSystemId": "myopenmhzsystem",
      "apiKey": "api-KEY"
    }
  ]
}
```
instead of having the plugin info commingled with system configuration data.

This should also directly address https://github.com/TrunkRecorder/trunk-recorder/issues/1087.

### DIFF
--- a/docs/CONFIGURE.md
+++ b/docs/CONFIGURE.md
@@ -314,20 +314,20 @@ By default, Trunk Recorder will record the call from the first site to receive t
 
 | Key     | Required | Default Value | Type                 | Description                                                  |
 | ------- | :------: | ------------- | -------------------- | ------------------------------------------------------------ |
-| library |    ✓     |               | string               | the name of the library that contains the plugin.            |
-| name    |    ✓     |               | string               | the name of the plugin. This name is used to find the `<name>_plugin_new` method that creates a new instance of the plugin. |
-| enabled |          | true          | **true** / **false** | control whether a configured plugin is enabled or disabled   |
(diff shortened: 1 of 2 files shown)


robotastic:
Great idea! This looks like a nice, clean upgrade!

Issue #1034, opened 2025-08-02
Crash with TR-5


lilhoser:
I guess we closed my last ticket too soon.  This looks like a different crash, though.

```
[2025-07-25 10:12:46.018569] (error)   [whiteoakmt-nbradley] Retuning to Control Channel: 769.606250 MHz
[2025-07-25 10:12:46.018810] (info)   	 - System Source 0 - Min Freq: 769.507625 MHz Max Freq: 771.523625 MHz
terminate reached from thread id: 7ebf984806c0Got unknown exception
0x7ec3af92d93d: (gr::tagged_stream_block::check_topology(int, int)+0x142d)
0x7ec3af0bb0da:[2025-07-25 10:12:46.025068] (info)   	 Xlating Channelizer decimator - freq_xlating taps: 79 Decim: 4 Resampled Rate: 24571.4 Lowpass Taps: 85
[2025-07-25 10:12:46.025253] (info)   	 Channelizer ARB - Symbol Rate: 24000 Resampled Rate: 24571.4 ARB Rate: 0.976744 ARB Taps: 223 BW: 0.390698 TB: 0.195349
 (std::rethrow_exception(std::__exception_ptr::exception_ptr)+0x8a)
0x7ec3af0a58e6: (__cxa_call_terminate+0x33)
0x7ec3af0ba8ba: (__gxx_personality_v0+0x31a)
0x7ec3af3f4b06: (_Unwind_GetTextRelBase+0x2c16)
0x7ec3af3f55cd: (_Unwind_Resume+0x12d)
0x7ec3afc17a5c: -- error: unable to obtain symbol name for this frame
0x7ec3b0a9d78d: (osmosdr::device::find(osmosdr::device_t const&)+0x5add)
0x7ec3af8f2db2: (gr::block_executor::~block_executor()+0x242)
0x7ec3af932965: (gr::prefs::save()+0x675)
0x7ec3af9329e4: (gr::prefs::save()+0x6f4)
0x7ec3afc2599f: (boost::detail::make_ready_at_thread_exit(boost::shared_ptr<boost::detail::shared_state_base>)+0x36f)
0x7ec3aec9caa4: (pthread_condattr_setpshared+0x684)
0x7ec3aed29c3c: (__clone+0x24c)
Aborted (core dumped)
```

gofaster:
[...]

Issue #1032, opened 2025-07-31
Crashing. Possibly due to poor signal, but not sure.


ancker010:
Getting crashes randomly. Here's a paste of the logs.

```
[2025-07-31 22:19:47.019410] (error)   [starcom21peso] Retuning to Control Channel: 852.412500 MHz
[2025-07-31 22:19:47.019668] (info)   	 - System Source 0 - Min Freq: 850.957000 MHz Max Freq: 853.257000 MHz
[2025-07-31 22:19:47.034783] (info)   	 Xlating Channelizer decimator - freq_xlating taps: 91 Decim: 4 Resampled Rate: 24000 Lowpass Taps: 83
rtlsdr_read_async returned with -2
rtlsdr_read_async returned with -2
terminate reached from thread id: 76facd6006c0Got unknown exception
terminate reached from thread id: 76f8834006c0Got unknown exception
0x76fbc268293d:0x76fbc268293d: (gr::tagged_stream_block::check_topology(int, int)+0x142d)
0x76fbc20640da: (gr::tagged_stream_block::check_topology(int, int)+0x142d)
0x76fbc20640da: (std::rethrow_exception(std::__exception_ptr::exception_ptr)+0x8a)
0x76fbc204e8e6: (std::rethrow_exception(std::__exception_ptr::exception_ptr)+0x8a)
0x76fbc204e8e6: (__cxa_call_terminate+0x33)
0x76fbc20638ba: (__cxa_call_terminate+0x33)
0x76fbc20638ba: (__gxx_personality_v0+0x31a)
0x76fbc1eb2b06: (__gxx_personality_v0+0x31a)
0x76fbc1eb2b06: (_Unwind_GetTextRelBase+0x2c16)
0x76fbc1eb35cd:[2025-07-31 22:19:47.130680] (error)   [starcom21peso]	 Control Channel Message Decode Rate: 1/sec, count:  4
 (_Unwind_Resume+0x12d)
0x76fbc25c4a5c: (_Unwind_GetTextRelBase+0x2c16)
0x76fbc1eb35cd: -- error: unable to obtain symbol name for this frame
0x76fbc2c8d78d: (_Unwind_Resume+0x12d)
0x76fbc25c4a5c: (osmosdr::device::find(osmosdr::device_t const&)+0x5add)
0x76fbc2647db2: -- error: unable to obtain symbol name for this frame
0x76fbc2c8d78d: (gr::block_executor::~block_executor()+0x242)
[...]

Issue #997, opened 2025-02-01
Mystery segmentation fault: Raspberry Pi 5


connervieira:
I'm running the latest version of Trunk Recorder as of the time of writing, installed via the documentation at <https://trunkrecorder.com/docs/Install/INSTALL-PI>

I'm using a Raspberry Pi 5 (6.6.74+rpt-rpi-2712) with two RTL-SDR v4.

I'm using a known-good configuration file from a previous installation (which worked reliably for weeks at a time). Here's the configuration contents:
```json
{
    "ver": 2,
    "sources": [{
        "center": 855059375,
        "rate": 2400000,
        "error": 2000,
        "gain": 50,
        "debugRecorders": 0,
        "digitalRecorders": 4,
        "driver": "osmosdr",
        "device": "rtl=0"
    }, { 
        "center": 861350000,
        "rate": 2400000,
        "error": 2000,
        "gain": 50,
        "debugRecorders": 0,
        "digitalRecorders": 4,
        "driver": "osmosdr",
        "device": "rtl=1"
    }],
    "systems": [{
        "control_channels": [855212500, 856037500, 860787500, 860912500],
        "type": "p25",
        "shortName": "OhioMARCS",
        "talkgroupsFile": "./talkgroup.csv",
        "modulation": "qpsk"
    }]
}
```

`trunk-record` starts just fine, and begins scanning the configured frequencies, but it will abruptly seg-fault anywhere from 10 seconds to a couple minutes after starting. Here's an example of the console output during such a scenario:
```
[2025-01-31 23:28:47.020475] (info)   	 - System Source 0 - Min Freq: 853.909375 MHz Max Freq: 856.209375 MHz
[2025-01-31 23:28:47.406096] (info)   	 Xlating Channelizer decimator - freq_xlating taps: 91 Decim: 4 Resampled Rate: 24000 Lowpass Taps: 83
fix_cc :warning: Block (fix_cc0) max output buffer set to 10240 instead of requested 8192
fix_cc :warning: Block (fix_cc1) max output buffer set to 8192 instead of requested 10240
Allocating 15 zero-copy buffers
Allocating 15 zero-copy buffers
[2025-01-31 23:28:47.485722] (error)   [OhioMARCS]	 Control Channel Message Decode Rate: 0/sec, count:  1
[2025-01-31 23:28:50.020621] (error)   [OhioMARCS] Retuning to Control Channel: 856.037500 MHz
[...]

Issue #980, opened 2024-10-27
Segmentation fault (core dumped) ... and getting discouraged


thedanielhunt:
Been working in earnest on my build with two RTL-SDRs, and the system keeps dumping out. I know many will say "turn off AGC" but that seems to be the only way to get signals captured. Anyone got any guidance on what I'm doing wrong?

```
{

  "ver": 2,
  "frequencyFormat": "mhz",
  "sources": [
  {
      "center": 853244000,
      "rate": 1512000,
      "gain": 7.7,
      "digitalRecorders": 16,
      "analogRecorders": 0,
      "driver": "osmosdr",
      "device": "rtl=0",
      "agc": true,
      "ppm": -1.5
    },
      {
      "center": 851732000,
      "rate": 1512000,
      "gain": 7.7,
      "digitalRecorders": 16,
      "analogRecorders": 0,
      "driver": "osmosdr",
      "device": "rtl=1",
      "agc": true,
      "ppm": -3.0
    }
  ],
  "systems": [
    {
      "shortName": "srrcs",
      "talkgroupsFile": "talkgroups.csv",
      "talkgroupDisplayFormat": "id_tag",
      "modulation": "qpsk",
      "squelch": -70,
      "control_channels": [
        853450000,
        852350000,
        853187500,
        853875000
      ],
      "channels": [
        851700000,
        851050000,
        851262500,
        851325000,
        851387500,
        851437500,
        851600000,
        851625000,
        851650000,
        851675000,
        851750000,
        851800000,
        851825000,
        851850000,
        852075000,
        852250000,
        852300000,
        852462500,
        852575000,
        852687500,
        852737500,
        853162500,
        853262500,
        853612500,
        853712500,
        853800000,
        853900000
      ],
      "type": "p25",
      "hideUnknownTalkgroups": false,
      "recordUnknown": false,
      "hideEncrypted": true
    }
  ],
  "captureDir": "recordings",
  "logDir": "logs"
}
```

thedanielhunt:
And this what the system is doing when it dumps out: 

```
[2024-10-27 15:51:37.250481] (info)   [srrcs]	31C	TG:       3565 (                      -)	Freq: 851.825000 MHz	Not Recording: TG not in Talkgroup File 
[...]

PR #1090, opened 2026-02-03
Fix P25 Control retune crash

mymegabyte:
Change tb->stop() and tb->start() to tb->lock() and tb->unlock, as they were prior to the v5.0 commit. This appears to the cause of  crashes when returning the control channel as referenced in issues #1034 #997 #1032 #972 #980.

### DIFF
--- a/trunk-recorder/monitor_systems.cc
+++ b/trunk-recorder/monitor_systems.cc
@@ -716,9 +716,6 @@ void retune_system(System *sys, gr::top_block_sptr &tb, std::vector<Source *> &s
         if (system->get_system_type() == "smartnet") {
           system->set_source(source);
           // We must lock the flow graph in order to disconnect and reconnect blocks
-          // ( We have gone back and forth on whether this should be lock/unlock or stop/wait/start.
-          //   If there are unexplained issues around control channel tuning, we should look at alternet
-          //   approaches. See PR #1090 )
           tb->lock();
           tb->disconnect(current_source->get_src_block(), 0, system->smartnet_trunking, 0);
           system->smartnet_trunking = smartnet_impl::make(control_channel_freq, source->get_center(), source->get_rate(), system->get_msg_queue(), system->get_sys_num());
@@ -728,6 +725,9 @@ void retune_system(System *sys, gr::top_block_sptr &tb, std::vector<Source *> &s
         } else if (system->get_system_type() == "p25") {
           system->set_source(source);
           // We must lock the flow graph in order to disconnect and reconnect blocks
+          // ( We have gone back and forth on whether this should be lock/unlock or stop/wait/start.
+          //   If there are unexplained issues around control channel tuning, we should look at alternet
+          //   approaches. See PR #1090 )
           tb->lock();
           tb->disconnect(current_source->get_src_block(), 0, system->p25_trunking, 0);
           system->p25_trunking = make_p25_trunking(control_channel_freq, source->get_center(), source->get_rate(), system->get_msg_queue(), system->get_qpsk_mod(), system->get_sys_num());


ot0tot:
I can confirm this change fixed my issue with trunk recorder crashing with error `rtlsdr_read_async returned with -2`.

robotastic:
I am game to give it a try - I remember we switched it for stop() for a reason... but I truly can't remember

taclane:
Aside from the flowgraph differences between lock() and stop(), there does seem to be some other factor that hasn't been totally identified.

I periodically have a source of interference that temporarily swamps one of my control channels, and t-r happily retunes across three or four RTL devices searching for a live channel until it lands back where it started.  So, even using stop() appears to work fine under some circumstances; we just haven't figured out if there's something else to it like:
>  t-r will retune crash if it's immediately after startup, but can cycle as long as it takes if the system successfully was in operation prior to the issue

mymegabyte:
It's likely a race condition. My quick research before trying this fix was that the recommended path was lock/unlock. But if we needed to keep stop/start for some reason there needs to be a tb->wait after the stop to prevent the race condition. I didn't try that since I saw it used to be lock/unlock. I can try changing it back and adding the wait to see if that's also valid if your prefer to try to honor the switch.

taclane:
I think ultimately returning to `lock()`/`unlock() `is the gnuradio-correct answer.  

But testing just now, i wasn't able to generate a segfault by forcing a retune on startup (channels out of order), or during runtime (interference event), with:
- `stop()`/`start()`
- `stop()`/`wait()`/`start()`
- `lock()`/`unlock()`

There's got to be some other t-r configuration or system or gnuradio factor.  It used to be something I could easily reproduce, but it's just a problem I totally stopped having.

mymegabyte:
Just for completeness I reconfigured back to my setup that triggered the crashes (directional antenna misaligned to source and lowered the gain), which results in frequent warnings about control channel message rate being under 10 followed by retune cycles to different control channels. 
lock() --> unlock(): Previously confirmed to work w/o crash in this setup (didn't re-test)
stop() --> start(): Took 30-45 minutes but eventually reproduced the crash
stop() -> wait() -> start(): Been running for about 6 hours, no crash.

So I think if the original reason for trying stop/start comes to surface, adding the wait would likely be a valid fix instead of lock/unlock.

taclane:
> So I think if the original reason for trying stop/start comes to surface, adding the wait would likely be a valid fix instead of lock/unlock.

I was scrolling back in time through the discussions on discord, and I don't think there was much reason other than it was discovered the retune event was becoming a failure point, so maybe trying a different way to adjust the flowgraph would help.

Well over a year ago, there was some chat about switching back or adding the wait(), but it's possible it just didn't go anywhere because the reports weren't as prevalent by then, and it was often solved by ensuring the active control channel was first in the channel list.

It probably should just to back to lock/unlock, but it would still be great to understand why it's been a bigger problem for some users than others.  It might just be something simple like the types/rates of devices, or having a fast-enough computer that stop() is able to finish the work loops before the p25 control channel decoder gets torn down and replaced.

robotastic:
agreed - it is a bit weird, but it is likely some race condition that will only get triggered with specific conditions. I am good to switch back - but I will leave a note in the code and a link to this PR, so we can pull it back up.

robotastic:
I will merge this in tomorrow and add the note

robotastic:
I totally missed that this was only a change for P25 and the we were doing that for smartnet already - this must have been a bad change that snuck in. I added a note just in case.

taclane:
> I totally missed that this was only a change for P25 and the we were doing that for smartnet already - this must have been a bad change that snuck in. I added a note just in case.

Last I recall when I had a core dump to look through, it's not that the switch that was the issue, the problem seemed to be that the previous p25 control channel decoder can be torn down _too fast_.  stop() is an asynchronous call to wrap things up, but depending on whether it was "locked on" first, either the gardner or costas block may still be hard at work trying to recover a clock signal then the connected control channel block just disappears from the flow graph.

It's potentially just been a matter of coincidence that some systems can shut things down fast-enough in the execution gap between `tb->lock()` and `tb->disconnect()`.

PR #1091, opened 2026-02-07
Harris OTA alias support for P25P1/P2

taclane:
Following the implementation of Motorola aliases in #1074, this PR will decode Harris/TIA-102 radio aliases for P25 TDMA/FDMA systems.

Unlike Motorola aliases, Harris aliases are not presented with fully qualified radio IDs (wacn:sys:suid), so the associated unit must be inferred from the active speaker. To minimize asynchronous matches (alias fragments may be received _ahead_ of t-r processing transmission info), the TDMA/FDMA decoders are cacheing the last known unit ID with a timestamp, and will only create the SUID/alias entry if:
- Unit ID < 300ms old
- The transmission sink is currently processing a call with a matching talkgroup and unit ID

For better accuracy in TDMA P2 systems, this PR incorporates @tadscottsmith's changes from the [`dev/id-fix`](https://github.com/TrunkRecorder/trunk-recorder/tree/dev/id-fix) branch to increase the likelihood of t-r having the correct information from the transmitting unit.

In testing, the proposed changes to TDMA handling have been very effective for P2 systems. Utilizing `MAC_HANGTIME` to signal the end of a unit transmission is within the bounds of P25 specifications as it is a guaranteed signal when the PTT is released. This avoids missed transitions between units as a `MAC_END_PTT` or grant message may not always occur during back-to-back transmissions. These changes will work whether the system is following **Call Continuation on Control Channel**, or **Call Continuation on Voice Channel** procedures.

### DIFF
--- a/lib/op25_repeater/lib/p25p1_fdma.cc
+++ b/lib/op25_repeater/lib/p25p1_fdma.cc
@@ -40,6 +40,14 @@
 #include "rs.h"
 #include "p25_crypt_algs.h"
 #include "imbe_vocoder/imbe_vocoder.h"
+#include <sys/time.h>
+
+// Helper to get current time in milliseconds since epoch
+static uint64_t get_time_ms() {
+	struct timeval tv;
+	gettimeofday(&tv, NULL);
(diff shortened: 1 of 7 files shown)


robotastic:
This is looks great! Thanks @taclane for all of the work putting this together

PR #1095, opened 2026-02-26
feat: add MQTT status plugin to Docker image

LumenPrima:
## Summary

Include [tr-plugin-mqtt](https://github.com/TrunkRecorder/tr-plugin-mqtt) in the official Docker build so users get MQTT support out of the box without maintaining a separate image.

### Changes (3 additions to `Dockerfile`)

- **Builder stage**: add `libpaho-mqtt-dev` and `libpaho-mqttpp-dev` to `apt-get install`
- **Builder stage**: clone `tr-plugin-mqtt` into `user_plugins/` before `cmake`
- **Runtime stage**: add `libpaho-mqtt-dev` and `libpaho-mqttpp-dev` for the shared libraries

### Context

Several projects (tr-engine, rdio-scanner, etc.) depend on MQTT output from trunk-recorder. Currently users must build a custom Docker image or use a community fork to get MQTT support. This adds it to the official image with minimal footprint — the plugin is already maintained under the TrunkRecorder org.

### Testing

- Verified the same approach builds and runs successfully in our multi-arch image ([ghcr.io/lumenprima/trunk-recorder-mqtt](https://github.com/LumenPrima/tr-engine/blob/master/docker/trunk-recorder-mqtt/Dockerfile))


Dygear:
It's worth noting that there was some discussion on the discord about this. More than a few people have requested that the MQTT plugin is included in the default build. Currently the @TheGreatCodeholio maintains a docker image that has this enabled as well.

Issue #1089, opened 2026-01-27
Harsh noise on P2 transmissions following Vocoder change


SimonSheehan:
Following the vocoder change, on the occasional iffy decoded transmission I will end up with a sort of Decepticon sounding awful noise. Other users noted it as well. 

This is very much specific to the changed P2 vocoder - I was previously reverting that commit and could safely merge, but since the P2 files have changed a bit with some new features, I no longer can, thus the problem has become unavoidable.

I do **not** believe the softVocoder option is working either as it should with these changes, as having that enabled does not help. In theory, it should be reverting back to the old P2 vocoder. 

[`f40ab20`](https://github.com/TrunkRecorder/trunk-recorder/commit/f40ab202ceff6a48ddd48615b39b2340044e0554)is the last good commit before this occurred. Commit [`065501e` ](https://github.com/TrunkRecorder/trunk-recorder/commit/065501e7ab0d3901be0f653ab385953ad8d1369b) is when it began. 

[Sound can be found in the T-R discord at this link.](https://discord.com/channels/928800455444791306/928801674087251988/1465818657253888192)

rpdale:
In addition many P25 Phase II transmissions have some clicks in them, and the new vocoder automatically levels down to keep those clicks quieter, but that causes the voice audio to be too low to be usable.

bburkett52:
I have been experiencing a similar issue with pager tones followed by voice
announcements: tones are loud and the following voice is very low volume.
Next transmission of voice only is normal.

This is on a P25P2 system. 

I feed T-R into RDIO and no changes there have any affect.

SimonSheehan:
Seeing a few more people pop in to the Discord and mention this lately. If anyone is able to check into the logic of softVocoder using the "old" one, it would be great (it's definitely outside of my scope of knowledge sadly)

robotastic:
Can you give the `dev/vocoder-fix` branch a try? I used Claude to fix a few issues which may have been contributing to it.

SimonSheehan:
Running it now - Will report back after a day of using it most likely.

blantonl:
I'm also interested in the results of this fix.

[...]

PR #1096, opened 2026-03-03
Vocoder Fix

robotastic:
Based upon some LLM analysis, Claude did the following: Summary
#	Issue	Verdict	Action
1	Uninitialized error_history_sqrt	Real bug (High)	Fixed: initialized to 0.0 2	Uninitialized samples_buf	Real bug (High)	Fixed: added memset to zero buffer 3	uint16_t* type mismatch	Real bug (Medium)	Fixed: changed to int16_t* 4	uint32_t to int16_t truncation	Low risk, defensive	Fixed: added & 0xFFFF mask 5	Missing default in fr_type switch	Real (Medium)	Fixed: initialized fr_type = FT_4V_0 6	crc12 returns 0 on overflow	Not practical	Skipped — not an audio issue 7	msg_len uninitialized for 0x01/0x21	False positive — b1b2 is always 0 for those opcodes	Added defensive init to 0 anyway 8	Loop variable shadowing	Not a bug	Skipped — cosmetic only The two highest-impact fixes for audio quality are Issue 1 (spike detection was comparing against garbage, making error tracking unreliable) and Issue 2 (stack garbage was being pushed directly to the audio output when frames had high error rates, producing pops/noise instead of silence).

### DIFF
--- a/lib/op25_repeater/lib/p25p1_fdma.cc
+++ b/lib/op25_repeater/lib/p25p1_fdma.cc
@@ -873,7 +873,7 @@ namespace gr {
 					}
 					double error_history_avg = error_history_total / error_history_len;
 
-					double error_history_sqrt;
+					double error_history_sqrt = 0.0;
 					for (int j=0; j<error_history_len; j++){
 						error_history_sqrt += pow((error_history[j] - error_history_avg), 2);
 					}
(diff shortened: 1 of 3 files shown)


robotastic:
Might address #1089

PR #1098, opened 2026-03-03
Have Plugins be able to update TR Config

robotastic:
There is probably a lot of fun that you could have with this. What do you think of the approach? Any setting missing?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -273,6 +273,7 @@ list(APPEND trunk_recorder_sources
   trunk-recorder/sources/iq_file_source.cc
   trunk-recorder/csv_helper.cc
   trunk-recorder/config.cc
+  trunk-recorder/config_service.cc
   trunk-recorder/setup_systems.cc
   trunk-recorder/monitor_systems.cc
   trunk-recorder/talkgroup.cc
--- a/trunk-recorder/config.cc
+++ b/trunk-recorder/config.cc
@@ -117,6 +117,9 @@ bool load_config(string config_file, Config &config, gr::top_block_sptr &tb, std
               << "byte position of error: " << e.byte << std::endl;
   }
 
+  // Store the raw parsed JSON so we can track which keys were explicitly set
+  config.loaded_json = data;
+
   try {
     // const std::string json_filename = "config.json";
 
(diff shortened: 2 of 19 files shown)


Dygear:
I didn't see a function that will do a config write out. I was thinking something along the lines of:

`config_write` / `config_save` that overwrites the current config file. What ever name you choose, if you decide to implement would be to allow you to write the config file to drive.

Param: `bool backup` - Default `false`, is no backup of config. `true` is backup the current config file before overwriting it.
Param: `string filename` - Default `NULL` so it saves the config with the current loaded filename + `.old`. Otherwise follows the same formatting rules as what is used for the new filename parser for the wav / json / m4a files.
`config_write(bool backup = false, ?string filename = NULL)`

taclane:
> I didn't see a function that will do a config write out. I was thinking something along the lines of:
> 
> `config_write` / `config_save` that overwrites the current config file. What ever name you choose, if you decide to implement would be to allow you to write the config file to drive.
> 
> Param: `bool backup` - Default `false`, is no backup of config. `true` is backup the current config file before overwriting it. Param: `string filename` - Default `NULL` so it saves the config with the current loaded filename + `.old`. Otherwise follows the same formatting rules as what is used for the new filename parser for the wav / json / m4a files. `config_write(bool backup = false, ?string filename = NULL)`

There's honestly nothing stopping a plugin from doing so already.  As long as the user invoking t-r has write access to the config file, a plugin can read/write to it using basic file operations.  I've been experimenting with this in the tr-web plugin to edit the config file over http and then send a signal to restart trunk-recorder.

Dygear:
I was thinking there should be a canonical way of doing it from within the system. So that plugin devs don't have the implement this perhaps poorly (I'm talking about myself here) many times. Namely to make sure that it respects config file rules with not setting a value if it's default in the json file, or having plugins have their own config file settings preserved by the system having the full view of everything.

robotastic:
good points @Dygear - it would be better to implement it once, correctly. It maybe a little tricky to not have the Default values trickle into the file.

robotastic:
added in the ability for plugins to save the current configuration to a config.json file. it tracks the current JSON values so you do not get all of the default values saved to the file.

Dygear:
This should make it in for the next version?

robotastic:
Probably handled by #1127

PR #1121, opened 2026-04-08
Enable simplestream plugin in CMakeLists

Dygear:
Re-enables the simple stream plugin being build by default in the CMake file.

Re discussion in #1107 to bring this back to default being built as well as multiple conversations in the discord.


robotastic:
Cool - let's add it back in

PR #1123, opened 2026-04-08
SimpleStream improvements and logging

jameson-dev:
Description for changes have been added to each commit. Happy for criticism on these changes 😄 

Here's a brief breakdown on the changes:

- Move streams, my_tcp_io_service, and max_tcp_index from file-scope globals into the Simple_Stream class as private members, preventing shared state between instances

- Fix BOOST_FOREACH loops to iterate by const reference rather than copying the struct by value on every audio packet

- Wrap TCP send() calls in try/catch to prevent a dropped connection from crashing the entire trunk-recorder process

- Log a warning when UDP send_to() fails silently so network errors are visible in logs

- Replace ip::address::from_string() with Boost.Asio resolvers for both UDP and TCP connections, allowing hostnames to be used in the config in addition to literal IP addresses

**Note:** I haven't actually tested these changes at all not having a TR setup to test with at the moment, but I'm rather confident of the changes. I would definitely test these changes first.


Dygear:
I really wanted someone to make `udp://hostname.tld:8600` or `tcp://hostname.tld:8600` possible. We got the hostname part, so that's cool. That way it's a much tighter config statement.

jameson-dev:
Something like this https://github.com/TrunkRecorder/trunk-recorder/pull/1123/commits/318c58ce84714ca4d6db3624752aba405444a280, @Dygear?

Dygear:
Exactly!

robotastic:
Is anyone able to give these a test? I don't have SimpleStream setup

Dygear:
I'll put it in on my next update cycle.

robotastic:
I resolved the conflicts and merged it in!

PR #1125, opened 2026-04-10
Audio path fixes

TheGreatCodeholio:
fix: correct ffmpeg concat path resolution and add debug WAV option
                                                                                                                                                                                                                                   
  Summary         
                                                                                                                                                                                                                                   
  - Fix broken audio rendering for calls with relative temp directories. The ffmpeg concat demuxer resolves paths inside a concat list file relative to the concat file's own directory, not the process working directory.        
  Previously the concat file was written to the capture directory (./audio/SCPD/2026/4/10/) while the transmission filenames inside it were CWD-relative (./temp/SCPD/...), producing a chimera path like
  ./audio/SCPD/2026/4/10/./temp/SCPD/... that does not exist. The concat file is now written to the same directory as the transmission files (the temp dir), with entries as bare filenames — ffmpeg resolves them correctly in    
  place.          
  - Remove the vestigial raw_filename field. Call_Data_t::raw_filename was never written to disk by any code path; it existed only as a naming stem for the (now-relocated) concat file and as a no-op in cleanup. It has been
  removed from the struct, the assignment in create_base_filename, and both cleanup branches in remove_call_files.                                                                                                                 
  - Add audio_postprocess.debug_wav option. When enabled, a .debug.wav file is saved alongside normal call output in the capture directory. It is produced via an ffmpeg stream copy — no filtering, resampling, or loudness
  normalization — preserving the exact sample rate and bit depth of the original recordings. Useful for comparing processed output against the raw source or diagnosing post-processing artifacts. The debug file is retained even 
  when audioArchive is false; it is only removed on plugin failure when archiveFilesOnFailure is also false.
                                                                                                                                                                                                                                   
  Test plan       

  - Confirm calls with a relative tempDir (e.g. ./temp) render successfully and produce valid .wav/.m4a output — no No such file or directory errors from ffmpeg                                                                   
  - Confirm calls with an absolute tempDir (e.g. /dev/shm) are unaffected
  - Confirm no .raw.wav files appear in the capture directory                                                                                                                                                                      
  - Enable debug_wav: true and confirm a .debug.wav appears in the capture directory alongside the normal call files                                                                                                               
  - Confirm the .debug.wav is unmodified audio (correct sample rate, no filtering) compared to the source transmissions                                                                                                            
  - Confirm .debug.wav is retained when audioArchive: false                                                                                                                                                                        
  - Confirm .debug.wav is removed when a call exhausts retries with archiveFilesOnFailure: false

### DIFF
--- a/docs/CONFIGURE.md
+++ b/docs/CONFIGURE.md
@@ -324,7 +324,7 @@ Each system can optionally define an `audio_postprocess` object to control clean
 ```json
 "audio_postprocess": {
 "enabled": false,
-"debug_wav": false,
+"outputRawAudio": false,
 "highpass_hz": 0,
 "lowpass_hz": 0,
 "bandreject_hz": 0,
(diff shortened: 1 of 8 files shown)


Dygear:
I test this out when the system isn't in heavy use. (So 4am again.)

robotastic:
Is debug_wav the clearest name for it? Would flat or raw or unprocessed be a better way to describe it?

TheGreatCodeholio:
> Is debug_wav the clearest name for it? Would flat or raw or unprocessed be a better way to describe it?

There could be a better name or description yes. debug_wav sounds more like a log thing than a audio thing. The name came from the idea it was a file for audio debugging purposes.

Maybe `outputRawWav` or `outputRawAudio`?

Current output
`call_info.debug_filename  = stem + ".debug.wav";`

TheGreatCodeholio:
renamed to outputRawAudio and changed the outputfile to `stem + ".raw.wav";` Updated all variable and function names related to be clear and concise

robotastic:
ok - recovered from vacation.

PR #1126, opened 2026-04-11
Add P25 Call Alert (TSBK 0x1F) decoding

hoosierscanner:
## Summary

Decodes P25 Call Alert messages (TSBK opcode 0x1F) from the control channel. Previously this was a stub that only logged a debug message and discarded the message.

- Extracts source unit ID from bits 16–39 and target unit ID from bits 40–63
- Sets `message_type = CALL_ALERT`, `message.source`, and `message.talkgroup`
- Logs at info level: `tsbk1f\tCall Alert\tSource Unit: X\tTarget Unit: Y`
- Adds `CALL_ALERT = 18` to the `MessageType` enum in `parser.h`
- Adds `case CALL_ALERT:` to the message dispatch switch in `monitor_systems.cc`

## Bit layout

From the P25 spec (confirmed via OP25 reference implementation):
```
args = dst << 24 | src   →   bits[16:39] = src, bits[40:63] = dst
```

## Test

Verified on a live P25 system — Call Alerts appear correctly in logs with accurate source and target unit IDs.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CALL-ALERT-PATCH.md
+++ b/CALL-ALERT-PATCH.md
@@ -1,0 +1,53 @@
+# P25 Call Alert (TSBK 0x1F) Decoding
+
+## Summary
+
+Adds proper decoding of P25 TSBK opcode 0x1F (Call Alert) to trunk-recorder.
+Previously this was a stub that only logged a debug message and discarded the message.
+Now it extracts source/destination unit IDs, sets the message type, and flows through
+the plugin system like all other unit events.
+
+## Changes
+
+### `trunk-recorder/systems/parser.h`
+Added `CALL_ALERT = 18` to the `MessageType` enum.
+
+### `trunk-recorder/systems/p25_parser.cc`
+Replaced the stub with real decoding:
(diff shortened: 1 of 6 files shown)


hoosierscanner:
Someone who actually codes will need to review this, all of the code is AI generated. Worked with what I found on the web about how Call Alerts are formatted to guide it. This PR only shows the CA in the log info level. 
Background, call alerts are sent over the system to alert another radio. On my local system they are used to activate relays on a base radio as a poor-man's firehouse alerting system. I have also seen them used on systems to alert ERs to incoming patient reports, or from dispatch to raise units. 
The next goal would be to send these through the MQTT feed so something can be done with the data.

robotastic:
Cool - This general approach makes sense! Let me see if I can clean up this PR a little this weekend and then merge it in. The overall intent seems pretty straightforward and would be great to add.

robotastic:
I have a cleaned up version of this on #1143

PR #1143, opened 2026-05-30
Add P25 Call Alert (TSBK 0x1F) decoding with plugin hook

robotastic:
Picks up #1126 from @hoosierscanner and finishes the wiring that was
described in the original patch notes but not actually included.

## What the original PR did

- Added `CALL_ALERT = 18` to `MessageType` (parser.h)
- Decoded TSBK 0x1F in `p25_parser.cc` — extracts source unit (bits
  16–39) and target unit (bits 40–63), sets `message.message_type`,
  `message.source`, `message.talkgroup`, logs at info level
- Added a `case CALL_ALERT:` to the dispatch switch in
  `monitor_systems.cc` (empty `break`)

## What this PR adds on top

The `CALL-ALERT-PATCH.md` in the original PR claimed it added a
`unit_call_alert` plugin hook, but those changes weren't actually in
the diff — the switch case was a no-op so plugins never saw the
event. Fixed here:

- `Plugin_Api::unit_call_alert(System*, long source_id, long talkgroup)`
  no-op default in `plugin_api.h` — backward-compatible with existing
  plugins
- `plugman_unit_call_alert(...)` declared in `plugin_manager.h` and
  implemented in `plugin_manager.cc`, matching the
  `plugman_unit_answer_request` pattern
- `unit_call_alert()` helper in `monitor_systems.cc`, and the
  `CALL_ALERT` case now calls it so plugins actually receive the event

Also drops three files from the original PR that don't belong in the
tree:

- `CALL-ALERT-PATCH.md` — developer-facing patch notes, not user docs
- `Dockerfile.call-alert` — builder-specific image, not a project asset
- `patches/simplestream-dangling-pointer.patch` — that fix already
  landed in #1107

## Test plan

- [x] `cmake .. && make trunk-recorder` builds cleanly
- [ ] Run against a live P25 system, confirm `tsbk1f Call Alert` lines
      appear with correct source/target unit IDs
- [ ] Confirm a plugin that overrides `unit_call_alert` receives the
      event

Credits: original decoding work by @hoosierscanner in #1126.


robotastic:
Let's do this!

PR #1147, opened 2026-06-26
Add P25 voice decryption support

lorenzolrom:
Forked from https://github.com/blantonl/trunk-recorder/tree/add-p25-decryption.

Adds P25 voice decryption support for AES-256, DES-OFB, and ADP/RC4.

This adds an encryption key config path, loads configured keys into the OP25 frame assembler when a P25 recorder starts, and wires the OP25 Phase 1/Phase 2 voice paths through the appropriate decryptor before vocoding audio.

It also fixes the encrypted Phase 1 audio path so successfully decrypted audio is emitted instead of being muted, propagates Phase 1 encryption status JSON back to the recorder, and keeps encrypted calls with configured keys from being skipped or deleted as undecryptable encrypted traffic.

Testing:
Tested locally on macOS with a full CMake build. Also tested with live DES-OFB P25 traffic via OpenMHz; audio that previously started garbled and then muted now decodes correctly.

Notes:
This branch was rebuilt cleanly from current TrunkRecorder master for PR review and excludes unrelated fork/OTA-alias history.

### DIFF
--- a/examples/encryption_keys.json
+++ b/examples/encryption_keys.json
@@ -1,0 +1,19 @@
+{
+  "keys": [
+    {
+      "keyid": "0x0001",
+      "algid": "0x84",
+      "key": "0123456789ABCDEF0123456789ABCDEF0123456789ABCDEF0123456789ABCDEF"
+    },
+    {
+      "keyid": "0x0002",
+      "algid": "0x81",
+      "key": "0123456789ABCDEF"
+    },
+    {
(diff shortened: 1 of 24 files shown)


robotastic:
This is really cool! I have no way of easily testing this - I would love for others to take a look. #1112 is an attempt at the same, but I don't think it finished. @blantonl can you take a look at this?

PR #1150, opened 2026-07-16
Prevent flowgraph shutdown hangs

EricTendian:
## The issue I ran into

I had an Airspy stop delivering samples. Trunk Recorder noticed the stalled source, concluded the active calls, finished the pending uploads, and started shutting down. The last message it logged was `stopping flow graph`, though, and the process never actually exited.

That left the container alive but unable to do any useful work. Since Docker still saw a running process, its restart policy could not bring the recorder back automatically.

I traced the hang to GNU Radio flowgraph shutdown. The Airspy source block can have a worker waiting for more samples after RX has already stopped. In that state, `top_block::stop()` does not necessarily wake the worker, so `top_block::wait()` can wait forever for it to finish.

## What I changed

I wrapped flowgraph shutdown in a 10-second watchdog. Normal shutdown still follows the same graceful `stop()` and `wait()` path. If GNU Radio does not finish within the timeout, Trunk Recorder logs a fatal error and exits with the exit code it was already going to return. That lets Docker or another process supervisor restart it instead of leaving a dead recorder running indefinitely.

I also added separate log messages after the stop request and after all worker threads have stopped. If this happens with another source or block, the logs will make it clearer which part of shutdown is stuck.

## Validation

- I built Trunk Recorder and all bundled plugins successfully in a clean Ubuntu 24.04 Docker builder image.
- `git diff --check` passes.

I could not replay the original USB/SDR failure because the affected live container had already been recreated.

### DIFF
--- a/trunk-recorder/main.cc
+++ b/trunk-recorder/main.cc
@@ -22,15 +22,19 @@
 #include <cassert>
 #include <chrono>
 #include <cmath>
+#include <condition_variable>
+#include <cstdlib>
 #include <exception>
 #include <fstream>
 #include <iostream>
+#include <mutex>
 #include <signal.h>
 #include <sstream>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string>
 #include <sys/time.h>
+#include <thread>
 #include <time.h>
 #include <unistd.h>
 #include <utility>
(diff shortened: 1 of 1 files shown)


SimonSheehan:
This has been a long term issue from time to time for me, would love to see it implemented.

SimonSheehan:
Been running this for a few days and seems to be working correctly, no shutdown hangs and systemd is being allowed to restart the service.

PR #1157, opened 2026-08-10
Add optional control-channel source affinity

### LINKED ISSUE #1156: Add opt-in control-channel source affinity for multi-source configurations
## Problem

When a trunked system's decode rate falls low, Trunk Recorder cycles through its configured control channels. If the next channel is outside the currently assigned source but covered by another source, `retune_system` moves the control-channel decoder to that source and rebuilds its portion of the shared GNU Radio flowgraph under a top-block lock.

Cross-source scanning is useful and should remain the default. In a multi-system deployment, however, repeated cross-source rebuilds from one marginal system were correlated with counter resets and brief decode interruptions on an otherwise stable system sharing the same top block. One observed rebuild also stalled in the graph restart path. This is related to the across-SDR behavior previously discussed in #417, but the requested feature is narrower: allow operators who deliberately cover every valid control channel on one source to prevent that system from moving between sources.

## Proposed behavior

Add an optional per-system `sourceAffinity` setting:

- Default to `false`, preserving today's cross-source scanning behavior.
- When `true`, allow normal control-channel retunes within the source selected at startup.
- When the next control channel falls outside that source, skip the retune and emit a clear warning instead of rebuilding the decoder on another source.
- Do not affect voice-recorder source selection.

This option is only appropriate when the assigned source covers every control channel the system may legitimately use; the documentation should state that requirement explicitly.

## Acceptance criteria

- Existing behavior is unchanged when the setting is absent or `false`.
- Same-source control-channel retunes continue when enabled.
- Cross-source control-channel retunes are blocked when enabled.
- A blocked retune produces a clear warning.
- Voice-recorder source selection remains unchanged.
- Configuration documentation and focused tests cover the behavior.

lilhoser:
## Summary

Adds an opt-in per-system `sourceAffinity` setting for trunked systems. When enabled, Trunk Recorder keeps that system's control-channel decoder on its startup-assigned source: same-source control-channel retunes continue normally, while an alternate outside the assigned source is skipped with a warning.

The setting defaults to `false`, so existing cross-source scanning behavior remains unchanged. Voice-recorder source selection is not affected.

## Why

Moving a control-channel decoder between sources requires disconnecting and rebuilding that decoder within the shared GNU Radio top block. In a multi-system field deployment, repeated cross-source rebuilds from one marginal system correlated with resets and brief decode interruptions on another stable system, and one rebuild stalled in the graph restart path. Operators who intentionally place every valid control channel for a site within one source need a way to prevent an unusable out-of-range candidate from causing those shared-graph rebuilds.

This is an opt-in safety boundary, not a replacement for cross-source scanning. It must only be enabled when the assigned source covers every control channel the system may use.

## Changes

- Parse and document the optional `sourceAffinity` system setting.
- Preserve same-source retunes when affinity is enabled.
- Block out-of-range cross-source control-channel retunes with a specific warning.
- Leave the default retune path and voice-recorder source selection unchanged.
- Add a focused decision test for all enabled/disabled and covered/uncovered cases.

## Validation

- Native Ubuntu 24.04 Release build completed successfully, including bundled plugins.
- `ctest --output-on-failure`: 1/1 passed.

Closes #1156


taclane:
As an alternative, would this be better addressed by some kind of `"controlExplicit": true` setting, and only ever let-t-r use the listed frequencies in the config?

This would address a secondary issue (t-r will absorb any control channels it learns from system update messages, which may not be desirable sometimes), and minimize the need for an affinity setting where it latches onto the first device by just letting the user have full control of where it attempts to tune.

